Issue #2, opened 2026-07-21
IngeTrazo (free SketchUp-alternative modeler) would like to contribute — roadmap, legacy formats & governance?


tuxiasumari:
Hi! First, thank you for OpenSKP — a clean-room, MIT, cross-platform `.skp`
parser is exactly the missing piece for the free/libre 3D ecosystem.

I maintain [IngeTrazo](https://github.com/tuxiasumari/ingetrazo), a GPL-3.0,
Linux-first 3D modeler (a free SketchUp alternative aimed at civil engineering /
architecture in Latin America, with a BIM→IFC bridge). Today we open `.skp`
files by shelling out to Trimble's proprietary `SketchUpAPI.dll` through Wine —
it works, but it's a proprietary, Windows-only, offline-hostile dependency we'd
love to **replace with a pure-Python parser like OpenSKP**. I've already built a
pluggable import seam so OpenSKP can drop in as the preferred backend, with the
DLL path kept only as a fallback.

I want to **contribute substantially and long-term**, not just file wishlist
issues. Before I dive in, three questions to align:

1. **Legacy formats.** Our users bring models from *any* SketchUp version.
   OpenSKP currently targets the VFF container (2021+). Is decoding the older
   **MFC binary format (v8–v20)** on the roadmap, or out of scope? That range
   matters a lot for real-world adoption.
2. **Fidelity gaps.** Where do you feel the parser is weakest today (textured
   materials, per-corner UVs, component/group hierarchy + instance transforms,
   layers/scenes)? I'd like to pick a slice to own — I have deep experience with
   texture UV mapping and component instancing from IngeTrazo's DAE importer.
3. **License & governance.** Confirming MIT stays MIT, and how you like
   contributions to flow (PR conventions, direction, any CLA).

**What I can bring that's uncommon:** a **differential validation harness +
real-world `.skp` corpus across versions**. Because we already run the Trimble
SDK (as a black box), I can generate ground-truth output for a given `.skp` and
**diff it against OpenSKP's parse** (geometry, materials, hierarchy) to pinpoint
discrepancies — a fast feedback loop toward "as faithful as the Blender
importer." This is strictly **black-box output comparison** (feed file in,
compare results), never DLL decompilation, keeping the clean-room lineage
intact — the same "observed `.skp` files + their COLLADA exports" methodology
the format work already relies on.

Happy to start with whatever's most useful to you: the diff harness, sample
files (with permission), or a specific fidelity fix. Thanks again for building
[...]

PR #3, opened 2026-07-21
feat(python): expose Material.id and materials_by_id for face-material resolution

tuxiasumari:
## What

Exposes the face → material join on the public API:

- **`Material.id: Optional[int]`** — the TLV material ID this material is known by (`None` when geometry never references it).
- **`SkpModel.materials_by_id: Dict[int, Material]`** — the full join table for `Face.material_id`. Several TLV IDs may alias one `Material`; each resolves here.

## Why

`Face.material_id` carries the numeric IDs from the TLV stream (e.g. `29491`), but `Material` exposed only `name/color/transparency` — no ID to join on. So a consumer of the public API could extract geometry *and* materials, yet had no way to tell **which face has which colour**. The join already existed, but only inside the internal exporter (`_core.build_scene`'s `material_id_to_name` + name-then-folder lookup); this PR surfaces that same resolution on `SkpFile.parse()`.

Context: this is the first concrete gap we hit wiring OpenSKP into [IngeTrazo](https://github.com/tuxiasumari/ingetrazo) as its pure-Python `.skp` importer (see #2) — geometry parses with an exact bounding box vs. our SketchUp-SDK oracle, but no per-face colours could be applied.

## How

In `SkpFile.parse()`, after converting materials, walk `parsed["material_id_to_name"]` and resolve each name through `parsed["materials"]` then `parsed["materials_by_folder"]` — the identical fallback the internal exporter uses — attaching the first ID seen to `Material.id` and every ID to `materials_by_id`.

Both fields are new with defaults → fully backwards-compatible.

## Validation

- **Real file (SketchUp 2022):** all **19/19** distinct `Face.material_id` values in the model resolve to their `Material` (correct name + RGB spot-checked).
- **4 new tests** in `tests/test_parser.py` (`TestMaterialIdJoin`): basic join, folder-alias resolution, unresolvable ID skipped, `id` defaults to `None`. They stub `full_parse`, so no `.skp` fixture is needed — same approach as the existing suite. Full suite: **39 passed**.
- CHANGELOG entry added under `[Unreleased]`.

Happy to adjust naming/placement to your taste — and if you'd rather expose a resolver method instead of the dict, glad to rework it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **Python**: `Material.id` and `SkpModel.materials_by_id` — expose the TLV
+  material IDs that `Face.material_id` references, so callers can resolve a
+  face's material (colour/transparency) from the public API. Previously the
+  join existed only inside the internal exporter.
+
 ## [0.2.0] — 2026-06-18
 
 ### Added
--- a/packages/python/src/openskp/model.py
+++ b/packages/python/src/openskp/model.py
@@ -127,11 +127,18 @@ class Material:
         color: RGBA colour tuple ``(r, g, b, a)`` with each value in 0–255.
         transparency: Opacity factor where ``0.0`` is fully transparent and
             ``1.0`` is fully opaque.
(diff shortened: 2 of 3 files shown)


iamahsanmehmood (approved):
Approved.

Reviewed the full diff and verified locally: 39/39 tests pass, `ruff check` clean, fully backwards-compatible (new defaulted fields only — `Material.id`, `SkpModel.materials_by_id`). The `id(mat_data)` join is sound because `_core` stores the same dict object in both `materials` and `materials_by_folder`. Nice work exposing the face->material join on the public API.

PR #4, opened 2026-07-21
feat(python): extract material textures (image bytes + tile size)

tuxiasumari:
## What

Extracts material **textures** — the second gap we hit integrating OpenSKP into IngeTrazo (#2, follows #3):

- **`Texture` dataclass** — `filename`, tile `width`/`height` (**inches**, SketchUp's internal unit), raw image `data` bytes, and a `save()` helper.
- **`Material.texture: Optional[Texture]`** — `None` for plain-colour materials. Backwards-compatible (new defaulted field only).

## Why

A material with `hasTexture="1"` carries a `<mat:texture>` block in its `material.xml`, and the image itself sits right next to the XML inside the embedded ZIP (`materials/<folder>/<image>`). Neither was surfaced, so a textured material parsed as its solid colour only — no way to render or export it faithfully.

The `xScale`/`yScale` attributes are the real-world tile size (how many inches one repetition of the image covers), which is exactly what a consumer needs to map the texture without per-face UVs (SketchUp's default projected-texture behaviour).

## How

`_core._extract_texture` reads the `<mat:texture>` element and the image bytes from the material's ZIP folder. One robustness detail found in real files: the stored image name can differ from `textureFilename` (observed: `"..._Safety.jpg"` in the XML vs `"..._Saftey.jpg"` on disk) — the folder's non-XML sibling is used as fallback.

## Validation

- **Real file (SketchUp 2022):** 2/2 textured materials extract with correct JPEG bytes (`FFD8` magic) and tile sizes (24×24 in and 6×6 in) — including the one that needs the sibling fallback.
- **4 new tests**, including an **end-to-end parse of a synthetic in-memory `.skp`** (header marker + embedded ZIP) that exercises the real `_core` extraction path — no fixture file needed. Full suite: **39 passed**.

Branched independently from `main` so it can merge before/after #3 in any order. Together with #3, a consumer can now resolve `Face.material_id → Material → colour or texture` — full material fidelity from the public API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **Python**: texture extraction — `Material.texture` (`Texture` dataclass:
+  `filename`, tile `width`/`height` in inches, raw image `data` bytes,
+  `save()` helper). Images are read from the material's folder inside the
+  embedded ZIP, with a sibling fallback when the stored image name differs
+  from `textureFilename`.
+
 ## [0.2.0] — 2026-06-18
 
 ### Added
--- a/packages/python/src/openskp/_core.py
+++ b/packages/python/src/openskp/_core.py
@@ -362,6 +362,60 @@ def extract_dynamic_properties(d007):
     return properties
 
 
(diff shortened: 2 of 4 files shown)


iamahsanmehmood (approved):
Approved.

Verified locally: 39/39 tests pass, `ruff` clean, backwards-compatible. Extraction is safe — it only reads entries that exist in the archive namelist, so no path-traversal.

Note for merge coordination: this work is fully contained in #13, which extends it with colorized-copy support. We'll most likely merge #13 and close this one to avoid a double-merge — approving here to record that the texture-extraction code itself passed review. Thank you.

PR #5, opened 2026-07-21
feat(python): expose instance-level materials (Instance.material_id)

tuxiasumari:
## What

Exposes **instance-level materials**: `Instance.material_id` — the material painted onto a component instance itself. Third in the series (#3 face materials, #4 textures); with this one, full material fidelity is reachable from the public API.

## Why

In SketchUp it's very common to paint a whole component rather than its faces ("paint the component"): the faces inside keep material `None` and **inherit** the instance's material. The instance node (`6419`) carries that material in the same `D007`/`D107` structure faces use — but the parser ignored it, so every instance-painted component came out uncoloured and consumers had no way to resolve the inheritance.

Found on a real SketchUp 2026 file: **24 of 274 instances carry a material** (granite pergolas, wood floors, paver walks) — ~444 m² of painted surface that parsed as unpainted.

## How

- `_core`: read `D007`/`D107` under the `6419` instance node into the builder's instance dict as `material_id`.
- `model.py`: `Instance.material_id: Optional[int]` (default `None`), with the inheritance rule documented — faces whose own `material_id` is `None` inherit it, as the official SDK resolves on export.

Backwards-compatible (new defaulted field only).

## Validation

- **3 new tests**, including an end-to-end TLV extraction: a synthetic `6419` node carrying `D007`/`D107` built byte-by-byte and run through the real `parse_tlv_recursive` + `_extract_geometry_from_nodes` — no fixture file needed. Full suite: **38 passed**.
- Real-file spot check (SU2026): all 24 instance materials resolve to known material names via `material_id_to_name`.

Branched independently from `main` — merges in any order relative to #3/#4.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **Python**: `Instance.material_id` — the material painted onto a component
+  instance itself (SketchUp's "paint the component", the same `D007`/`D107`
+  structure faces use). Faces with no material of their own inherit it;
+  consumers can now resolve that inheritance like the official SDK does.
+
 ## [0.2.0] — 2026-06-18
 
 ### Added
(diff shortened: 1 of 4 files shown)


iamahsanmehmood (approved):
Approved.

Verified locally: 38/38 tests pass, `ruff` clean, backwards-compatible (new defaulted `Instance.material_id`). The read mirrors the existing instance-material resolution in `build_scene` (the same D007/D107 structure), so it's consistent with proven code. Independent of #3/#4 — no conflict expected.

PR #6, opened 2026-07-21
feat(python): decode per-face texture mapping (Face.uv_transform)

tuxiasumari:
## What

Decodes **per-face texture mapping** — `Face.uv_transform` / `uv_transform_back`. Fourth in the series (#3 face materials, #4 textures, #5 instance materials); this is the one that makes positioned and photo-fitted textures renderable.

## Why

A texture that has been positioned or photo-fitted with SketchUp's pins stores a per-face mapping the parser ignored. Without it, consumers can only apply the material's default projection — a photo-fitted waving-flag mesh, for example, renders as a single solid colour (the corner of the image every face happens to sample).

## The reverse engineering

**Location** (TLV archaeology on real SU2022/2026 files): under the face's entity-info blob,
`D007 → DC05 → DD05 → B136 → B236 → 1027 → 1127 (front) / 1227 (back) → 1327 → 1527` — a 3×3 row-major f64 matrix.

**Semantics** (calibrated with a controlled experiment: a SketchUp file with three 1×1 m squares — texture untouched / rotated 90° / pin-distorted — plus SDK-exported ground-truth UVs): the stored matrix maps **texture space → face plane**. UVs are recovered by inverting it:

```
xr = normalize(Z × n);  yr = n × xr        # plane basis from the face normal, inches
uvq = [p·xr, p·yr, 1] @ inv(M)             # row-vector convention
u = uvq[0]/uvq[2]/tile_w ;  v = uvq[1]/uvq[2]/tile_h
```

- Untouched texture → no matrix → default projection `u = (p·xr)/tile_w`.
- Distorted (4-pin) mappings are **projective** (`uvq[2] ≠ 1`) and come out exact under the same recipe.

**Validation:** < 0.0003 UV error on the controlled squares (including the distorted one), and **rms < 1e-5 over 150 photo-fitted triangles** of a real waving-flag mesh, both against SDK-exported ground truth.

## API

`Face.uv_transform` / `Face.uv_transform_back` (9-tuples, default `None`), with the full recipe documented on the dataclass — so any consumer (renderer, exporter) can apply it without re-deriving the convention. Backwards-compatible (new defaulted fields only).

## Tests

4 new tests with byte-built `DC05` blobs (no fixture files): front-only extraction, front+back, untouched → `None`, and the recipe reproducing known ground-truth UVs. Full suite: **39 passed**.

Branched independently from `main` — merges in any order relative to #3/#4/#5.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **Python**: per-face texture mapping — `Face.uv_transform` /
+  `uv_transform_back` (the 3×3 matrix a positioned / photo-fitted texture
+  stores per face; SketchUp's texture pins). Includes the decoded recipe to
+  turn it into UVs (plane basis from the normal, then
+  `[x, y, 1] @ inv(M) / tile`), calibrated against SDK-exported ground
+  truth to < 0.001 UV error, including projective (4-pin distorted)
+  mappings.
+
 ## [0.2.0] — 2026-06-18
 
 ### Added
--- a/packages/python/src/openskp/_core.py
+++ b/packages/python/src/openskp/_core.py
@@ -193,6 +193,66 @@ def extract_entity_id(node):
     return None
 
 
+def _tlv_flat(payload):
+    """Walk a raw payload as a flat TLV sequence; returns [(tag_hex, body)]."""
+    pos = 0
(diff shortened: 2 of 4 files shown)


iamahsanmehmood (approved):
Approved.

Verified locally: 39/39 tests pass, `ruff` clean, backwards-compatible. The library side only stores the 3x3 matrix (defensive, bounds-checked TLV walk; returns None at every missing level and checks `len == 72` before unpacking), so runtime risk is minimal. The documented UV recipe is validated against ground truth in the tests. Solid reverse-engineering write-up too.

Merge heads-up: this rewrites the same `builder.faces[f_id] = {...}` statement and `Face` dataclass as #11, so whichever lands second will need a quick rebase to union the new keys (`uv_transform`/`uv_transform_back` + `back_material_id`).

PR #7, opened 2026-07-21
fix(python): decode entity names as UTF-8, not ASCII-with-ignore

tuxiasumari:
## What

One-word fix with outsized impact: entity names decode as **UTF-8** instead of `ascii, errors='ignore'`.

## Why

SketchUp stores names UTF-8 encoded. Decoding as ASCII and *ignoring* the non-ASCII bytes silently corrupts any accented name — `"cópia"` → `"cpia"`, `"Diseño"` → `"Diseo"`. That's a guaranteed hit for files authored in Spanish, Portuguese, French, German…

And it's worse than cosmetics: **the TLV material name is the join key to the XML material files.** A corrupted name never matches its XML entry, so the material becomes unresolvable from geometry.

**Real-world case** (SketchUp 2026 plaza, authored in Spanish): the tree-foliage material `Celtis_australis_s cópia` decoded as `..._s cpia`, failed the join, and every tree crown in the model rendered colourless — we chased this as a "missing Image entities" bug for a while before finding it was an encoding bug all along.

## How

All six decode sites in `_core` (material names, layer names, definition names, instance names, dynamic-property keys/values) switch to `utf-8, errors='replace'`. ASCII names are byte-identical under UTF-8, so nothing changes for English files.

## Validation

- Real file: `material_id_to_name[3319170]` now yields `'Celtis_australis_s cópia'`, matching its XML entry — the join resolves and the foliage gets its texture.
- New TLV-built test: an instance named `"Diseño de árbol"` survives extraction intact. Full suite: **36 passed**.
- CHANGELOG entry under `[Unreleased]` (Fixed).

Branched independently from `main` — merges in any order relative to #3/#4/#5/#6.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **Python**: entity names (materials, layers, definitions, instances,
+  dynamic properties) now decode as **UTF-8** instead of ASCII-with-ignore.
+  Dropping the non-ASCII bytes silently corrupted any accented name
+  ("cópia" → "cpia", "Diseño" → "Diseo") and — critically — broke the
+  material-name join between the TLV stream and the XML material files,
+  leaving those materials unresolvable from geometry.
+
 ## [0.2.0] — 2026-06-18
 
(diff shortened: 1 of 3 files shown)


iamahsanmehmood (approved):
Approved.

Verified locally: 36/36 tests pass, `ruff` clean. Confirmed all six decode sites are converted and no `decode('ascii'` remains in `src/`. Since UTF-8 is a strict superset of ASCII, English/ASCII files are byte-identical — zero regression — while accented names are now correct and the material-name join key is fixed. High-value, low-risk fix; this is the one we plan to land first.

PR #8, opened 2026-07-21
feat(python): parse Image entities (9013/401F placement + Definition.is_image)

tuxiasumari:
## What

Parses **Image entities** — a picture placed in the model as an object. Sixth in the series (#3–#7).

## Why

An Image entity was completely invisible: its quad definition parsed fine, but its **placement** wraps a standard `6419` instance node inside two image-specific containers (`9013` → `401F`) that weren't in `CONTAINER_TAGS` — so the inner instance stayed buried in an opaque payload, the definition looked "never placed", and the image never appeared in any consumer's output.

**Real-world case** (SketchUp 2026 plaza): two photo cut-outs placed as images — a statue and a bull — imported with no geometry at all.

## How

- `CONTAINER_TAGS += {'9013', '401F'}` — the wrapped `6419` now parses like any instance and the existing extraction picks it up with its placement matrix. That's the whole fix for placement: two tags.
- `Definition.is_image` — the definition backing an image carries TLV kind `8315 == 2` (ordinary definitions carry 0/1); exposed so consumers can give images special treatment (billboard cut-outs, image lists, etc.). Backwards-compatible (new defaulted field).

## Validation

- Real file: instance count 274 → **276**, and the two new instances' `ref_idx` match the image definitions exactly, with sane placement matrices (scale + upright rotation + translation).
- 2 new tests: a byte-built `9013/401F/6419` extraction through the real parser, and `is_image` mapping through `parse()`. Full suite: **37 passed**.

Branched independently from `main` — merges in any order relative to #3–#7.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **Python**: Image entities — a picture placed in the model as an object
+  now parses: its placement wraps a standard instance node inside the
+  image-specific `9013`/`401F` containers (previously opaque, so the image
+  definition looked "never placed"), and `Definition.is_image` flags the
+  single-quad definition backing it (TLV kind `8315 == 2`). Real-world
+  case: photo cut-out statues/animals placed as images imported with no
+  geometry at all.
+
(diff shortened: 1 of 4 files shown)


iamahsanmehmood (approved):
Approved.

Verified locally: 37/37 tests pass, `ruff` clean, backwards-compatible. Confirmed `9013`/`401F` aren't consumed as raw payload anywhere on `main`, so promoting them to container tags is safe — and `parse_tlv_recursive` is bounds-checked, so a mis-shaped payload can't corrupt output. `is_image` is a clean defaulted addition.

Merge heads-up: shares the `collect_defs` loop with #7 and #9, so a small rebase will be needed depending on merge order.

PR #9, opened 2026-07-21
feat(python): decode the "always faces camera" component behavior

tuxiasumari:
## What

Decodes the **"always face camera"** component behavior — `Definition.always_faces_camera`. Seventh in the series (#3–#8).

## Why

SketchUp's 2D people and cut-out components (the classic *Susan*) carry the "always face camera" behavior: viewers must rotate them to face the camera. The flag was undecoded, so a consumer had no way to know which components to billboard — they rendered as static flat planes seen edge-on from most angles.

## The reverse engineering

Found by differential comparison across a real file's definitions — three components known to face the camera (the 2D person "Susan" and two photo cut-out wrappers) against static ones (chairs, tree parts, groups):

```
Susan:          581B → 5D1B=1  5E1B=1     ← faces camera
cut-out wrap 1: 581B → 5D1B=1  5E1B=1     ← faces camera
cut-out wrap 2: 581B → 5D1B=1  5E1B=1     ← faces camera
chair/leaf/branch/groups: 5D1B=0 5E1B=0   ← static
```

Perfect separation: the definition's `581B` behavior block carries sub-TLV **`5D1B == 1`** exactly on face-camera components; its companion `5E1B` is "shadows face sun" (SketchUp's usual pairing).

## API

`Definition.always_faces_camera: bool` (default `False`), read in `collect_defs`. Backwards-compatible.

## Validation

- Real file (SU2026): the three known face-me definitions read `True`, everything else `False`.
- New synthetic end-to-end test: a byte-built `.skp` (header + ZIP with a hand-assembled `model.dat` containing two definitions, flag on/off) round-trips through `parse()`. Full suite: **37 passed**.
- CHANGELOG under `[Unreleased]`.

Branched independently from `main` — merges in any order relative to #3–#8.

PR #10, opened 2026-07-21
feat(python): parse style face colors (SkpModel.styles)

tuxiasumari:
## What

Parses the file's **style face colors** — `SkpModel.styles` with `Style(name, front_color, back_color)`. Eighth in the series (#3–#9).

## Why

Unpainted faces take their look from the file's style. Without it, a viewer shades them with its own defaults and diverges from what the author saw — back-face tints are often picked deliberately (e.g. to read as grass or paper).

## How

`styles/*/style.xml` in the embedded ZIP stores signed-int32 ARGB variants: item **4000** = default front face color, **4001** = back face color. Decoded to RGB 0-255 tuples.

## Validation

Synthetic end-to-end `.skp` test (byte-built ZIP with a hand-written style.xml) through `parse()`; real-file check on a SU2026 model returns both colors. Full suite: **36 passed**. CHANGELOG under [Unreleased]. Backwards-compatible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **Python**: styles — `SkpModel.styles` (`Style`: name, `front_color`,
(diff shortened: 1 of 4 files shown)


iamahsanmehmood (approved):
Approved.

Verified locally: 36/36 tests pass, `ruff` clean, backwards-compatible. Purely additive (`Style` dataclass + `SkpModel.styles`), defensive XML parsing (guards `ParseError`/`ValueError`), and the signed-int32 ARGB -> RGB decode is correct. Independent of the geometry/material PRs.

PR #11, opened 2026-07-22
feat(python): expose the face's back-side material (Face.back_material_id)

tuxiasumari:
## What

Exposes the material of a face's **back side** — `Face.back_material_id`, read from the `AF0D` child of the face node. Ninth in the series (#3–#10).

## Why

SketchUp faces have independent front and back materials. A face painted **only on its back** is common in real files: authors paint the *visible* side of a downward-facing cap (e.g. garden beds modeled as caps whose front faces down — the user paints the up-facing back with grass). Without this field such faces parse as unpainted and render colourless.

**Real-world case** (SU2026 plaza): 4 garden-bed faces carry the grass material only in `AF0D` — front `D107` absent. In SketchUp they read green from above; consumers of the parser showed them unpainted.

## How

`AF0D` (a direct child of `AC0D`, sibling of `D007`) holds the back material id — verified by tracing every raw occurrence of a known material id through the TLV tree (5× `D107` fronts + 4× `AF0D` backs in the case file, matching SketchUp's Entity Info exactly: "Front: default / Back: grass").

## Validation

Byte-built `AC0D` node test through the real extractor (front stays `None`, back resolves). Full suite: **37 passed**. CHANGELOG under [Unreleased]. Backwards-compatible (new defaulted field).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **Python**: `Face.back_material_id` — the material of a face's BACK side
+  (the `AF0D` child of the face node). A face painted only on its back is
+  common when the author paints the visible side of a downward-facing cap;
+  without this field such faces looked unpainted.
+
 ## [0.2.0] — 2026-06-18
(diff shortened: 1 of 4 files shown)


iamahsanmehmood (approved):
Approved.

Verified locally: 37/37 tests pass, `ruff` clean, backwards-compatible (new defaulted `Face.back_material_id`). Correct `AF0D` read with defensive guards.

Merge heads-up: rewrites the same `builder.faces[f_id] = {...}` statement and `Face` dataclass as #6 — whichever lands second needs a quick rebase to union the added keys.

PR #12, opened 2026-07-22
fix(python): material transparency must honour useTrans

tuxiasumari:
## What

`Material.transparency` now honours the `useTrans` flag — tenth in the series (#3–#11).

## Why

In `material.xml`, the `trans` attribute **only applies when `useTrans="1"`**. Otherwise it's a leftover default. Ignoring the flag makes most materials read as 50% transparent (`trans` defaults) and some as fully invisible (`trans="0"` with `useTrans="0"` — a real chain-link fence material in the case file). Real-world spread in one SU2017-era file re-saved as VFF: ~70 materials carried `trans` values, only 3 of them with `useTrans="1"` (the actual translucent glass).

## How

One conditional in the material XML parse: `transparency = trans if useTrans == "1" else 1.0`.

## Validation

Two synthetic end-to-end `.skp` tests (byte-built ZIP + hand-written material.xml): `useTrans="1"` applies the value; `trans="0" useTrans="0"` stays opaque. Full suite: **37 passed**. CHANGELOG under [Unreleased].


tuxiasumari:
Pushed a follow-up commit: the stored `trans` value is a **transparency** (0 = opaque), not an opacity — SketchUp's library "Translucent Glass Blue" renders at 70% opacity and stores `trans="0.3"`. The exposed value is now `1 - trans` (clamped). The classic MFC format stores the same transparency f64 gated by a use-flag byte, and the two formats agree under this reading (see #14).

iamahsanmehmood (dismissed):
Verified locally: 37/37 tests pass, `ruff` clean. `trans` should only apply under `useTrans="1"`, and exposing opacity as `1 - trans` matches the documented meaning of `Material.transparency` (0.0 transparent .. 1.0 opaque). Good catch on the `trans="0"`/`useTrans="0"` invisible-material case.

One thing to flag for release (not a code issue): unlike the rest of the series, this changes the *value* of an existing field rather than adding a new one. After upgrading, `Material.transparency` returns different numbers for the same file — most materials move 0.5 -> 1.0, and translucent ones invert. GLB export is unaffected (face alpha is hard-coded to 255), but any consumer reading `Material.transparency` directly will see different results without changing their own code. So we'll ship this one deliberately, with a prominent CHANGELOG "Changed" note (and likely its own version) so downstream apps opt in knowingly. Approving the fix as correct.

iamahsanmehmood:
Pulled back the earlier review — the fix itself is correct, but the PR needs one addition before it's ready to ship.

**What's needed:** unlike the rest of the series, this one changes the *value* of an existing field rather than adding a new one — after this lands, `Material.transparency` returns different numbers for the same file (most materials move 0.5 → 1.0, translucent ones invert). That's the right fix, but it's silent right now: the CHANGELOG entry is filed under "Fixed" with no flag that existing consumers reading `Material.transparency` will see different values without changing their own code.

Concretely: please move/duplicate the entry under a "Changed" heading (or add an explicit note under "Fixed") that calls out the value change directly — something like "this changes the numeric value of `Material.transparency` for most files; audit any code that reads this field directly." That gives downstream consumers (including our own production apps) a clear signal to check before upgrading, instead of finding out from a diff.

Once that note is in, this is ready to go.

tuxiasumari:
Done 👍 Moved the entry under a **Changed** heading with an explicit ⚠️ value-change warning: it now spells out that `Material.transparency` returns different numeric values for the same file after this release (most `0.5 → 1.0`, translucent ones invert, e.g. Translucent Glass Blue `0.3 → 0.7`), and tells consumers to audit any code reading the field directly before upgrading.

iamahsanmehmood (dismissed):
Approved.

The CHANGELOG entry now clearly flags the value change under a Changed heading with an explicit warning — before/after numbers spelled out, audit instruction included. Verified locally: 37/37 tests pass, ruff clean.

iamahsanmehmood (approved):
Re-approving after updating the branch against main (conflict resolution only — CHANGELOG union + test file splice; no change to the useTrans logic itself, which was already reviewed and verified). All 13 CI checks pass.

iamahsanmehmood (approved):
Re-approving after the second branch update (PR #14 merged in between, requiring another sync — again just conflict resolution, no logic change to the useTrans fix itself). All 13 CI checks pass.

PR #13, opened 2026-07-22
Resolve shared texture bytes + expose colorized materials

tuxiasumari:
Builds on #4 (texture extraction) — the last commit is the one to review.

## What

Colourized copies of textured materials (the `[Name]1` materials SketchUp creates when you re-colour a textured material, `type="2"` in the material XML) store **no image of their own** — their `<mat:images>/<mat:image path>` points into the **source** material's folder (`materials/<other>/<image>`, sometimes prefixed `./`). Before this change, `Material.texture.data` came back `None` for every colourized material.

- `_extract_texture` now follows the `<mat:image path>` attribute (both archive-relative and `./`-prefixed forms) when the material's own folder holds no image, so the shared bytes resolve.
- `Material` gains `colorized` (`type == "2"`) and `colorize_type` (the `colorizeType` attribute: `0` = hue shift, `1` = tint) so viewers can re-tint the shared image toward `Material.color` the way SketchUp renders these copies.

## Why it matters

Real-world files use colourized materials heavily (e.g. a green-tinted `Fencing_Chain_Link`). Without the shared bytes the material silently loses its texture; with them + the flag a viewer can reproduce SketchUp's exact rendering. Validated against a real SketchUp 2021+ file where 4 colourized materials (`[Fencing Chain Link]1`, `[Wood Board Cork]1`, …) previously returned `data=None` and now resolve their images.

## Tests

`test_colorized_copy_shares_source_image` — synthetic .skp with a base material + a colourized copy whose image path points across folders; asserts the shared bytes, the `colorized` flag, `colorize_type` and the material colour. 40 tests green on the branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **Python**: texture extraction — `Material.texture` (`Texture` dataclass:
+  `filename`, tile `width`/`height` in inches, raw image `data` bytes,
+  `save()` helper). Images are read from the material's folder inside the
+  embedded ZIP, with a sibling fallback when the stored image name differs
+  from `textureFilename`.
+
 ## [0.2.0] — 2026-06-18
 
 ### Added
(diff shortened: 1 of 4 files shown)


iamahsanmehmood (approved):
Approved.

Verified locally: 40/40 tests pass, `ruff` clean, backwards-compatible. Shared-texture resolution checks candidates against the archive namelist before reading, so no path-traversal. This PR contains all of #4 plus colorized-copy support, so we plan to merge this one and close #4.

Two small asks before merge:
1. The CHANGELOG entry is copied from #4 and doesn't mention the colorized feature (the headline of this PR) — could you add a line for it?
2. Minor nit: `img_path.lstrip('./')` strips a character-set rather than a prefix; `removeprefix('./')` is more precise (harmless for `materials/...` paths, just tidier).

Great work resolving the shared-bytes case.

PR #14, opened 2026-07-22
feat: read classic (pre-2021) MFC-container .skp files

tuxiasumari:
Pre-VFF SketchUp files (SketchUp 2013–2020) are not ZIP containers: after the same UTF-16 header records, the body is **one uncompressed MFC `CArchive` object stream** with a single global store map. This PR adds `openskp/legacy.py`, a complete clean-room walker for that stream, and routes `full_parse()` to it on detection — `SkpFile.open(path).parse()` now handles both eras transparently, extending openskp's coverage to effectively every `.skp` in circulation.

## What the walker reads

- **MFC object layer**: null / new-class / class-ref / back-ref tags, the `FF7F` big-tag escape, per-file lazily-defined class slots, contextual learning for classes defined in the unwalked pre-model region, and slot-base bootstrap from the second material's class-ref tag.
- **Geometry**: the half-edge kernel (`CVertex`/`CEdge`/`CEdgeUse`/`CLoop`/`CFace`) with front/back materials, hole loops, per-face UV matrices (`CFaceTextureCoords`, including the per-side PROJECTED flag used by Add Location terrain drapes), arcs/curves, construction geometry, dimensions, texts and section planes.
- **Components**: definitions (incl. nested/inlined ones and the always-faces-camera behavior flag) and instances/groups with their 13-float transforms.
- **Materials & layers**: solid + textured materials (embedded JPEG/PNG bytes, applied tile size), opacity gated by the format's use-flag byte, colourized-copy detection, layer colours.

The walker is **self-checking**: every `CEdgeUse` carries a back-ref to its parent loop, which must resolve to the loop currently being parsed — any slot-numbering drift fails loudly instead of silently mis-linking topology.

## Validation

Cross-validated against the **same real-world models saved in both formats** (five architecture models authored in SketchUp 2016/2017/2018, re-saved as VFF with SketchUp Web, 130 KB–8 MB): exact face and edge counts, exact total surface area, and exact bounding boxes on all five, plus two more large v16 files parsed end-to-end. Version-specific record deltas (v16 has no persistent-id mask; v18 grows `CComponentDefinition`/`CComponentInstance`) are keyed off the header version.

The dict keys not yet consumed by `model.parse()` on main (`edge_flags`, `uv_projected`, `colorized`, UV matrices) pair with #3/#4/#6/#13 and my `edge-display-flags` PR; without them the reader still yields full geometry, materials and instances.

Unit tests included for container detection and the MFC string records (real-file tests would need binary fixtures; happy to add a small one if you want).


iamahsanmehmood (dismissed):
Verified locally: 42/42 tests pass, `ruff` clean, `legacy.py` imports fine. The routing is safe: `is_legacy` is triple-guarded (SketchUp magic + no `PK\x03\x04` ZIP in the first 0x100 bytes + `CVersionMap` marker), and the import is lazy inside `full_parse`, so the VFF path existing users depend on is provably untouched. A classic file that previously failed with "No ZIP container found" can now only improve. `full_parse_legacy` returns the exact same dict shape as `full_parse`, and it already emits the fields the rest of the series adds — nice forward-consistency.

One request before we actually merge/release this one: the ~880-line MFC walker currently has unit tests only for container detection and the string records, not the geometry decode. Since the repo is used by production apps, could you add a small real-file binary fixture (even one tiny v16/v17 model) that asserts face/edge counts + bounding box, the way your cross-validation does? That gives us a real regression guard for the decoder itself. A redacted/minimal fixture is totally fine. Approving the design now; we'll want that fixture in place before it ships.

iamahsanmehmood:
Pulled back the earlier review — the design and routing safety are solid, but there's one concrete thing needed before this can move forward.

**What's needed:** a real-file binary fixture test for the geometry decode itself. Right now the ~880-line MFC walker only has unit tests for container detection (`is_legacy`) and the string records — the actual geometry/topology decode (vertices, edges, faces, materials, instances) has no regression test against a real file. Since this repo backs production applications, we need that guard in place before it ships.

Concretely: add one small real `.skp` fixture (a tiny v16 or v17 model is fine — doesn't need to be complex, and a redacted/minimal one works) and a test that runs it through `full_parse_legacy` / `SkpFile.open().parse()`, asserting face count, edge count, and bounding box — the same kind of check your PR description mentions using for cross-validation.

Once that fixture test is in and green, this is ready to go.

tuxiasumari:
Added the real-file regression fixture you asked for 👍

`packages/python/tests/fixtures/capilla_quiroz_v17.skp` — a small real SketchUp 2017 model (v17.0.18899, ~212 KB), and `TestLegacyRealFile` in `test_parser.py` runs it through `SkpFile.open().parse()` and asserts version, face/edge/vertex counts (181 / 515 / 335), material count, instance placement and the bounding box. The expected values were cross-validated against the same model re-saved as VFF by SketchUp Web (identical face/edge counts, area and bbox across the two formats). Exempted from the repo's `*.skp` ignore rule so it stays checked in.

Whole `test_parser.py` green locally (45 passed).

iamahsanmehmood (approved):
Approved.

The real-file fixture is exactly what this needed — a genuine SketchUp 2017 model with face/edge/vertex counts, material count, instance placement, and bounding box all asserted against it, cross-validated against the same model re-saved as VFF. Ran it myself: 3/3 new tests pass, 45/45 full suite, ruff clean. Great addition — this is a real regression guard for the decoder now, not just container detection.

PR #15, opened 2026-07-22
feat: expose per-edge display flags (Edge.soft/smooth/hidden)

tuxiasumari:
The edge record's `D007` container carries a `D307` flag byte: base `0x06`, plus `0x01` = **hidden**, `0x08|0x10` = **soft/smooth**. Established by correlating ~26,000 edges of real models against the same files in the classic MFC format, whose edge record stores the three flags as separate bytes: plain edges read `0x06`, hidden `0x07`, softened+smoothed `0x1E` across every sample.

`Edge` gains `soft`/`smooth`/`hidden` booleans (default `False`, so existing consumers are unaffected).

Why it matters for viewers/exporters: soft/smooth flags are what hide the facet lines of curved surfaces, and — just as important — their *absence* is what keeps author-drawn coplanar subdivisions visible (glass mullions, beam/column lines). Angle-based smoothing heuristics can't distinguish the two cases; the file's own flags can.

Pairs naturally with #14 (the classic MFC reader fills the same `builder.edge_flags` from its format's native flag bytes).

### DIFF
--- a/packages/python/src/openskp/_core.py
+++ b/packages/python/src/openskp/_core.py
@@ -158,6 +158,7 @@ class _GeometryBuilder:
     def __init__(self):
         self.vertices = {}
         self.edges = {}
+        self.edge_flags = {}      # edge id -> display flag byte (D307)
         self.faces = {}
         self.instances = []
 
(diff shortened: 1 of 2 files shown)


iamahsanmehmood (approved):
Approved.

Verified locally: 35/35 tests pass, `ruff` clean, backwards-compatible (new defaulted `Edge.soft/smooth/hidden`), and nicely defensive with `getattr(builder, "edge_flags", {})`. The bit interpretation (0x01 hidden, 0x08 soft, 0x10 smooth) is consistent with the cross-validation notes.

One request before we merge: this is the only PR in the series without its own test or CHANGELOG entry. Could you add a small decode test (e.g. 0x06 plain, 0x07 hidden, 0x1E soft+smooth) and a changelog line? Then it's good to go.

PR #17, opened 2026-07-23
fix(ts): decode entity, layer, and material names using utf-8 instead of ascii

nooraliqureshi:
### Description
Fixes an encoding issue in the TypeScript package (\packages/typescript\) where string names for entities, component definitions, layers, dynamic property keys/values, and materials were decoded using \TextDecoder('ascii')\ instead of \TextDecoder('utf-8')\.

When files contain non-ASCII characters or accented names (e.g., \cópia\, \Diseño\, or multilingual text), decoding with \scii\ silently corrupts or strips those characters.

### Changes Included
- Replaced \
ew TextDecoder('ascii')\ with \
ew TextDecoder('utf-8')\ across 6 locations:
  - \packages/typescript/src/geometry.ts\:
    - Component instance name decoding (\	ag 6519\ constraint)
    - Layer name decoding (\	ag 8D3C\ constraint)
    - Component definition name decoding (\	ag 7E15\ constraint)
    - Dynamic property key and value decoding (\	ags B636 & AD38\)
  - \packages/typescript/src/index.ts\:
    - Material name decoding (\	ag CC32\ constraint)

### Verification & Testing
- **Typecheck**: Ran \
pm run typecheck\ (\	sc --noEmit\) — 0 errors.
- **Unit Tests**: Ran \
pm run test\ (\itest run\) — all 12 tests passed.
- **Build**: Built CJS, ESM, and DTS bundles via \
pm run build\ (\	sup\) — build succeeded cleanly.


iamahsanmehmood (approved):
Approved.

Verified everything myself rather than relying on the PR description: ran `tsc --noEmit` (clean, 0 errors), `vitest run` (12/12 passed), and `tsup` build (CJS/ESM/DTS all succeeded). Also confirmed the actual bug independently — `new TextDecoder('ascii')` isn't 7-bit ASCII in JS, the WHATWG spec aliases that label to windows-1252, so decoding UTF-8 bytes for "cópia" with it produces mojibake ("cÃ³pia"), not truncation like Python's old bug. Switching to `'utf-8'` fixes it correctly. All 6 sites converted, confirmed via diff and a grep for any remaining `TextDecoder('ascii')` — none left.

Nice catch, and good timing given we just fixed the equivalent bug on the Python side (#7).

PR #30, opened 2026-08-06
Add C++17 OpenSKP implementation

### LINKED ISSUE #29: Support for C++
Hey folks,

Thanks for this great initiative! I've been hacking on parsing SKP files in C++ as well, but I think my efforts are more effective by contributing to this open-source initiative.

- The target language and ecosystem: C++ with the CMake build system.
- Your experience with that language: 10+ years of experience with C++.
- Estimated timeline: 1 week.

Regards,
Thomas

p.s. I'm unable to add the `platform` label to this issue.

thomasloockx:
Closes #29

## Summary

- add `packages/cpp`, an independent C++17 implementation for modern VFF/ZIP and legacy MFC SKP containers
- expose parsed model, scene baking, triangulation, transforms, structured errors, progress, and logging APIs
- provide an installable CMake package with static/shared builds and `OpenSkp::OpenSkp`
- add Google-style clang-format configuration with a 100-column limit and cross-platform CI
- document the C++ API, dependencies, architecture, and contributor workflow

## Dependencies

- miniz 3.1.2, pinned and compiled privately
- GoogleTest 1.17.0, fetched only when tests are enabled

## Validation

- 33 GoogleTest tests pass against modern and legacy fixtures
- AddressSanitizer and UndefinedBehaviorSanitizer suite passes
- clang-format check passes
- install and downstream `find_package(OpenSkp CONFIG REQUIRED)` consumer are covered by CI

The initial C++ port exposes GLB-ready scene data but intentionally does not include GLB, OBJ, or JSON serializers.


iamahsanmehmood:
Thanks for this — really solid work. The library itself builds clean and all 33 tests pass on every platform in the matrix (Linux/gcc, Linux/clang, macOS/clang, Windows/MSVC), plus ASan/UBSan and clang-format are clean too.

One failure to fix before merge: the "Configure installed-package consumer" step fails on **Windows only**:

```
CMake Error at CMakeLists.txt:3 (find_package):
  Could not find a package configuration file provided by "OpenSkp"
```

This isn't a problem with the library or its install rules — `OpenSkpConfig.cmake` does get installed to the expected `prefix/lib/cmake/OpenSkp/` path (visible earlier in the same log). The likely cause is in the workflow step itself: that step runs under Git Bash on the Windows runner, but is passed `-DCMAKE_PREFIX_PATH=${{ github.workspace }}/prefix`, which evaluates to a native Windows path like `D:\a\openskp\openskp/prefix`. Bash's own parser treats those backslashes as escape characters before `cmake` ever sees the argument, which corrupts the path.

Could you swap that one argument to a bash-safe form, e.g. `-DCMAKE_PREFIX_PATH="${GITHUB_WORKSPACE}/prefix"` (using the `GITHUB_WORKSPACE` env var, which Git Bash presents in POSIX form) instead of the `${{ github.workspace }}` expression? Once that consumer smoke test passes on Windows too, this is ready to go.

iamahsanmehmood (approved):
All CI green across the full matrix (Linux gcc/clang, macOS clang, Windows MSVC, ASan/UBSan, clang-format), and the implementation cross-validates exactly against the other four languages' fixtures on real files. Thanks for the fix and for the contribution.

Issue #28, opened 2026-08-06
Legacy MFC: definition-tail CThumbnail scan misses big-tag class-refs (slot > 32767)


thomazyujibaba:
## Summary

`readDefinition` fails with `definition tail: thumbnail not found` on legacy (MFC `CArchive`) `.skp` files whose object count pushes the `CThumbnail` class slot past **32767**.

The thumbnail *is* present in the file. The scanner just cannot recognise it, because it only handles the short class-ref encoding.

## Root cause

In `readDefinition`, the 96-byte scan that locates the definition-tail `CThumbnail` tests only the 16-bit class-ref form:

```js
const thumbSlot = ar.classSlot.get("CThumbnail");
if (thumbSlot !== void 0) {
  if (view.getUint16(p, true) === (32768 | thumbSlot)) {   // <-- here
    tpos = p;
    break;
  }
}
```

When `thumbSlot > 32767`, `32768 | thumbSlot` exceeds `0xFFFF`, so it can never equal a `getUint16` result. The comparison is dead code and the scan always falls through to the throw:

```
slot=278   -> (32768|slot)=33046 fitsU16=true
slot=65712 -> (32768|slot)=98480 fitsU16=false   // unreachable
```

MFC `CArchive` encodes large slot indices with a big-tag escape: `0x7FFF` followed by a `u32` carrying `0x80000000 | slot`. `readObject` already implements both forms:

```js
if (tag === 32767) {
  const big = r.u32();
  if (big & 2147483648) {
    return this.newOfClass(r, big & 2147483647, expect);
  }
  return this.backref(big, r);
}
```

The tail scanner was never taught the escape form.

## Evidence

Two real files, both legacy MFC, both authored in the same tool. Failure correlates with slot value, not with file content or encoding:

| File | Size | `CThumbnail` slot | Bytes at scan+43 | Result |
|---|---|---|---|---|
| A (works) | 3.2 MB | 278 | `16 81` = `0x8116` = `0x8000\|278` | parses, 30 defs |
| B (fails) | 36 MB | 65712 | `FF 7F B0 00 01 80` | throws at `0x1065a64` |

In file B, `FF 7F` is the `0x7FFF` escape and the following `u32` is `0x800100B0` = `0x80000000 | 65712` — a correctly encoded class-ref to the thumbnail. Every definition in both files places the marker at the same offset (`scan+43`), so the tail layout itself is consistent; only the encoding width differs.

Instrumenting the walker on file B:

```
[...]

PR #39, opened 2026-08-08
fix(python): recognise big-tag CThumbnail class-refs in the definition-tail scan (ports #28)

tuxiasumari:
This is the Python side of #28 — the same 16-bit-only check @thomazyujibaba found in the TypeScript scanner exists in `legacy.py`, as you noted in that issue. Rather than wait for the TS PR to land and port it over, here it is ready.

## The bug

```python
if struct.unpack_from('<H', r.data, p)[0] == 0x8000 | ar.class_slot['CThumbnail']:
```

Once the `CThumbnail` class slot passes 0x7FFF, `0x8000 | slot` cannot fit in 16 bits, the comparison is dead code, and every definition tail throws `thumbnail not found` — while `read_object` had decoded the big-tag escape (`0x7FFF` + u32 of `0x80000000 | slot`) all along.

## The fix

Both encodings now live in one helper, `_is_class_ref`, mirroring `read_object` exactly. Unit tests cover the short form, the escape form, the truncated-data edge, and the specific pre-fix failure (a big slot must never match the truncated 16-bit comparison).

## Verification

- The 36 MB repro from #28 (the office chair posted there) now parses: **7 definitions, 145 353 faces, 0 non-finite coordinates** — face count matching the TS fix's GLB triangle count exactly.
- Swept **64 other corpus files** with an order-stable SHA-256 fingerprint of every vertex coordinate: **all byte-identical** before and after the fix.
- `pytest`: 84 passed.

### DIFF
--- a/packages/python/src/openskp/legacy.py
+++ b/packages/python/src/openskp/legacy.py
@@ -201,6 +201,21 @@ class _Archive:
 
 # ── shared record blocks ─────────────────────────────────────────────────
 
+def _is_class_ref(data: bytes, p: int, slot: int) -> bool:
+    """True when the bytes at *p* are an MFC class-ref to class *slot*.
+
+    Mirrors both encodings that ``_Archive.read_object`` decodes: the short
+    16-bit form ``0x8000|slot`` and, for slots past 0x7FFF, the big-tag
+    escape ``0x7FFF`` followed by a u32 of ``0x80000000|slot``.
+    """
+    if slot <= 0x7FFF:
+        return (p + 2 <= len(data)
+                and struct.unpack_from('<H', data, p)[0] == 0x8000 | slot)
+    return (p + 6 <= len(data)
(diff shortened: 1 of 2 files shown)


iamahsanmehmood (approved):
Clean fix, and good catch tying it back to #28's TS root cause instead of re-deriving it. The `_is_class_ref` helper correctly mirrors both encodings `read_object` already handles (short form for slot <= 0x7FFF, big-tag escape otherwise) — verified against our own walker's equivalent dispatch logic.

Appreciate the dedicated unit tests here, especially `test_big_slot_never_matches_short_form` — that's exactly the case that was silently wrong before (truncated comparison that can never fit in 16 bits). All CI green.

Also confirmed: our C#/.NET port's definition-tail thumbnail scanner has the identical short-form-only check, so this same fix is needed there too — we'll port it in a follow-up.

Approving.

Issue #38, opened 2026-08-07
Several exceptions using C# implementation


japajoe:
Maybe I misunderstood, but I thought I should be able to open skp files made with Sketchup Make 2013. I can not get it to open any file. I have 2 scenarios:

- Trying to open a file made with Sketchup Make 2013 (it just contains a simple cube).
- Trying to open a file made with Sketchup 8, which then is saved as a Sketchup Make 2013 file.

Both files fail to open but each have their own type of exception.

Result of opening simple model saved with Sketchup Make 2013:
```
Unhandled exception. OpenSkp.SkpParseException: legacy .skp parse failed: definition list misaligned @0x1c571: ...00000794DCE6C8604EE7BE0A208F54E8 | 9974FFFEFF074400650072007200690063006B00FFFEFF984400650072007200... | stage=legacy_walk
 ---> OpenSkp.LegacyParseError: definition list misaligned @0x1c571: ...00000794DCE6C8604EE7BE0A208F54E8 | 9974FFFEFF074400650072007200690063006B00FFFEFF984400650072007200...
   at OpenSkp.LegacyReaders.ReadDefinition(Archive ar, LR r) in /home/user/Desktop/OpenSkp/src/OpenSkp/Legacy.cs:line 868
   at OpenSkp.Archive.NewObj(LR r, String name) in /home/user/Desktop/OpenSkp/src/OpenSkp/Legacy.cs:line 303
   at OpenSkp.Archive.NewOfClass(LR r, Int32 cslot, String expect) in /home/user/Desktop/OpenSkp/src/OpenSkp/Legacy.cs:line 292
   at OpenSkp.Archive.ReadObject(LR r, String expect) in /home/user/Desktop/OpenSkp/src/OpenSkp/Legacy.cs:line 271
   at OpenSkp.Legacy.Walk(Byte[] data) in /home/user/Desktop/OpenSkp/src/OpenSkp/Legacy.cs:line 1089
   at OpenSkp.Legacy.FullParseLegacy(Byte[] data, SkpParseOptions options) in /home/user/Desktop/OpenSkp/src/OpenSkp/Legacy.cs:line 1239
   --- End of inner exception stack trace ---
   at OpenSkp.Legacy.FullParseLegacy(Byte[] data, SkpParseOptions options) in /home/user/Desktop/OpenSkp/src/OpenSkp/Legacy.cs:line 1243
[...]

PR #40, opened 2026-08-08
fix(python): SketchUp 2013 instances have no trailing GUID — fixes both crashes in #38

tuxiasumari:
Root cause and fix for **both** failure modes in #38. They are one bug with two faces, and it is not C#-specific — every port that shares the legacy walker has it (verified against the Python one; the walker logic is the same in TS/.NET/Dart).

## The bug

`_read_instance` unconditionally consumes a 16-byte GUID after the instance name:

```python
xf = r.f64s(13)
name = r.utf16()
guid = r.raw(16)
```

That field only exists from **`CComponentInstance` schema 5** (and `CGroup` schema 1) on. **SketchUp 2013 writes `CComponentInstance` schema 4**, whose record ends at the name. So on 2013-era files every in-definition instance overruns the stream by exactly 16 bytes, and the definition tail (`nrel`, guid, name) is read misaligned. Both attached repros are format-version 13 (`TestSketchup8.skp` was re-saved as 2013, so it is v13 too), and both fail at the first definition that contains an instance — the template figure components ("Derrick", "Susan") that every SketchUp template ships:

| file | first in-def instance | symptom |
|---|---|---|
| TestSketchup2013.skp | `Derrick` figure | `definition list misaligned` (Python) / `LegacyParseError @0x1c571` (C#) |
| TestSketchup8.skp | same overrun, corrupts a later back-ref | `back-ref to unwalked slot` (Python) / `Nullable object must have a value` in `ReadEntityList` (C#) |

Worth knowing when triaging: those two signatures are not two bugs.

## The fix

The archive already reads and stores each class's MFC schema number — it just never consulted it. The instance reader now keys the trailing GUID on it: `CComponentInstance` ≥ 5, `CGroup` ≥ 1, unknown schema keeps today's behavior. (These thresholds match the schema ranges the independent Rust implementation pins for the same records.)

## Verification

- Both #38 repros parse with finite geometry: `TestSketchup2013.skp` → 3 definitions, 1 617 vertices; `TestSketchup8.skp` → 15 definitions, 39 274 faces, 26 286 vertices. Definition names come out as expected (`Rodeo#2`, `Derrick`, `Pillar`, `Mall`, `Susan`, …).
- Swept 63 other corpus files (v17 files heavy with instances included: 1 186-instance theater, 143-instance attributes file) with an order-stable SHA-256 fingerprint of every vertex coordinate: **all byte-identical** before and after.
- `pytest`: 80 passed.

The same schema gate should port directly to the TS/.NET/Dart walkers.

### DIFF
--- a/packages/python/src/openskp/legacy.py
+++ b/packages/python/src/openskp/legacy.py
@@ -125,6 +125,8 @@ class _Archive:
         self.r = _R(data)
         self.slots: Dict[int, tuple] = {}    # slot -> ('class'|'obj', name, value)
         self.class_slot: Dict[str, int] = {}
+        self.class_schema: Dict[str, int] = {}
+        self.current_class: Optional[str] = None  # class of the object being read
         self.next_slot = 0
         self.walk_base = 0            # slots below this are unwalked pre-model
         self.readers: Dict[str, Any] = {}
@@ -155,6 +157,7 @@ class _Archive:
             name = r.raw(namelen).decode('ascii')
             self.alloc(('class', name, schema))
             self.class_slot[name] = self.next_slot - 1
+            self.class_schema[name] = schema
             return self._new_obj(r, name)
         if tag & 0x8000:                       # class ref -> new object
             return self._new_of_class(r, tag & 0x7FFF, expect)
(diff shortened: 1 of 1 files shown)


iamahsanmehmood (approved):
Verified independently — ported this exact schema gate to the C# walker and ran it against both repro files plus 11 real production files (1.2 MB–591.7 MB):

- `TestSketchup2013.skp` → 3 definitions, 1,617 vertices
- `TestSketchup8.skp` → 15 definitions, 39,274 faces, 26,286 vertices

Both match your reported numbers exactly, including the component names (`Derrick`, `Rodeo#2`, `Pillar`, `Mall`, `Susan`, ...). Zero regressions on the 11-file real-world corpus. Root-cause analysis is solid and the fix is minimal and correctly scoped. Nice work tracking this down — the "one bug, two crash signatures" framing made triage a lot clearer.

Approving. C#/TS/Dart ports of the same schema gate will follow in separate PRs.

PR #41, opened 2026-08-08
fix(dotnet): SketchUp 2013 instances have no trailing GUID — fixes both crashes in #38

iamahsanmehmood:
Ports the schema-gated fix from #40 (Python) to the C# legacy walker. Same root cause, same fix — different language.

## The bug

`ReadInstance` unconditionally reads a 16-byte GUID after the instance name:

```csharp
var xf = r.F64s(13);
string name = r.Utf16();
var guid = r.Raw(16);
```

That field only exists from **`CComponentInstance` schema 5** (and `CGroup` schema 1) on. **SketchUp 2013 writes `CComponentInstance` schema 4**, whose record ends at the name. On 2013-era files every in-definition instance overruns the stream by exactly 16 bytes, and everything after it (the definition tail, subsequent backrefs) reads misaligned.

## The fix

The archive already tracked each class's registration slot (`ClassSlot`); this adds tracking of the schema number itself (`Archive.ClassSchema`) and which class is currently being read (`Archive.CurrentClass`), so `ReadInstance` can key the trailing GUID read on the actual schema instead of assuming it's always present — mirroring `#40`'s Python approach field-for-field.

## Verification

- Both #38 repro files parse with finite geometry:
  - `TestSketchup2013.skp` → 3 definitions (incl. root), 1,617 vertices
  - `TestSketchup8.skp` → 15 definitions (incl. root), 39,274 faces, 26,286 vertices
  - Both numbers match #40's reported Python results exactly, including component names (`Derrick`, `Rodeo#2`, `Pillar`, `Mall`, `Susan`, ...).
- Swept 11 real production files (1.2 MB–591.7 MB) with the existing regression harness: all parse identically before/after, zero regressions.
- Full `dotnet test` suite: 16 passed.

### DIFF
--- a/packages/dotnet/OpenSkp/Legacy.cs
+++ b/packages/dotnet/OpenSkp/Legacy.cs
@@ -214,6 +214,8 @@ namespace OpenSkp
         public LR R;
         public Dictionary<int, SlotEntry> Slots = new Dictionary<int, SlotEntry>();
         public Dictionary<string, int> ClassSlot = new Dictionary<string, int>();
+        public Dictionary<string, int> ClassSchema = new Dictionary<string, int>();
+        public string? CurrentClass;
         public int NextSlot;
         public int WalkBase;
         public Dictionary<string, LegacyReader> Readers = new Dictionary<string, LegacyReader>();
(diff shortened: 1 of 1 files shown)


thomasloockx:
As an aside, it might be worth taking those 2 example files (`TestSketchup2013.skp`) and (`TestSketchUp8.skp`) into the test suites for all languages to avoid future regressions.

iamahsanmehmood:
**Review:** schema tracking is threaded correctly through `ReadObject`/`NewObj`/`ReadInstance`, matches #40's Python semantics exactly (including the `schema == null` conservative-default fallback). CI green (1/1). Verified against both #38 repro files (exact match to #40's reported numbers) plus 11 real production files (1.2 MB–591.7 MB) with zero regressions. Merging with admin privileges since GitHub blocks self-approval on same-account PRs.

PR #42, opened 2026-08-08
fix(dotnet): recognise big-tag CThumbnail class-refs in the definition-tail scan (ports #28)

iamahsanmehmood:
Ports the fix from #39 (Python) / the original TS fix on #28 to the C# legacy walker. Same root cause, same fix, different language.

## The bug

```csharp
if (p + 2 <= r.Data.Length && Tlv.ReadU16(r.Data, p) == (0x8000 | thumbSlot))
```

Once the `CThumbnail` class slot passes `0x7FFF`, `0x8000 | slot` cannot fit in 16 bits, the comparison is dead code, and every definition tail throws `"thumbnail not found"` — while `ReadObject` had decoded the big-tag escape (`0x7FFF` + u32 of `0x80000000 | slot`) all along.

`CThumbnail`'s slot is the raw global object-slot number where its class was first registered — not a count of distinct class *types* — so any file where the first `CThumbnail` happens to be encountered after ~32,767 other objects have already been allocated hits this, independent of file size or complexity in any simple sense.

## The fix

Both encodings now live in one helper, `LegacyBytes.IsClassRef`, mirroring `ReadObject` exactly (matches Python's `_is_class_ref` in #39). Unit tests cover the short form, the escape form, the truncated-data edge, and the specific pre-fix failure (a big slot must never match the truncated 16-bit comparison).

## Verification

- Swept the same 11-file real-world corpus used for prior regression testing (1.2 MB–591.7 MB): all parse identically before/after, zero regressions.
- Full `dotnet test` suite: 20 passed (16 existing + 4 new).

Credit to @thomazyujibaba for the original TS report/repro (#28) and @tuxiasumari for confirming and porting to Python (#39).

### DIFF
--- a/packages/dotnet/OpenSkp.Tests/LegacyTests.cs
+++ b/packages/dotnet/OpenSkp.Tests/LegacyTests.cs
@@ -114,4 +114,49 @@ namespace OpenSkp.Tests
             Assert.True(Legacy.IsLegacy(bytes));
         }
     }
+
+    /// <summary>
+    /// Both MFC class-ref encodings the definition-tail scanner must match.
+    /// Mirrors packages/python/tests/test_parser.py::TestLegacyClassRef
+    /// (openskp#28 / #39): once a class slot passes 0x7FFF the short 16-bit
+    /// form 0x8000|slot can't fit, so MFC escalates to the big-tag escape
+    /// (0x7FFF followed by a u32 of 0x80000000|slot).
+    /// </summary>
+    public class LegacyClassRefTests
+    {
+        [Fact]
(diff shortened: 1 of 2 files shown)


iamahsanmehmood:
**Review:** `IsClassRef` correctly mirrors both encodings `ReadObject` decodes, matches #39's Python `_is_class_ref`. Good test coverage including the truncated-short-form regression case (the exact bug this fixes). CI green (1/1). Zero regressions on the real-file corpus. Merging with admin privileges since GitHub blocks self-approval on same-account PRs.

PR #43, opened 2026-08-08
fix(typescript): SketchUp 2013 instances have no trailing GUID — fixes both crashes in #38

iamahsanmehmood:
Ports the schema-gated fix from #40 (Python) / #41 (.NET) to the TypeScript legacy walker. Same root cause, same fix, different language.

## The bug

`readInstance` unconditionally reads a 16-byte GUID after the instance name:

```ts
const xf = r.f64s(13);
const name = r.utf16();
const guid = r.raw(16);
```

That field only exists from **`CComponentInstance` schema 5** (and `CGroup` schema 1) on. **SketchUp 2013 writes `CComponentInstance` schema 4**, whose record ends at the name. On 2013-era files every in-definition instance overruns the stream by exactly 16 bytes.

## The fix

The archive already tracked each class's registration slot (`classSlot`); this adds tracking of the schema number itself (`Archive.classSchema`) and which class is currently being read (`Archive.currentClass`), so `readInstance` can key the trailing GUID read on the actual schema — mirroring #40's Python approach field-for-field.

## Verification

- Both #38 repro files parse (isolated build, this fix alone, without the separate #28/#39 class-ref fix): `TestSketchup2013.skp` → 2 definitions, `TestSketchup8.skp` → 14 definitions — matching #41's (.NET) numbers.
- Swept 9 of the same 11 real production files used for prior regression testing (1.2 MB–109 MB): all parse identically before/after, zero regressions. The remaining 2 files (280–592 MB) hit a pre-existing Node/V8 heap limit in this package's in-memory model representation, independent of this change — reproduces identically on unmodified main.
- Full `vitest` suite: 35 passed.

### DIFF
--- a/packages/typescript/src/legacy.ts
+++ b/packages/typescript/src/legacy.ts
@@ -174,6 +174,8 @@ class Archive {
   r: R;
   slots = new Map<number, SlotEntry>();
   classSlot = new Map<string, number>();
+  classSchema = new Map<string, number>();
+  currentClass: string | null = null;
   nextSlot = 0;
   walkBase = 0;
   readers: Record<string, (ar: Archive, r: R) => any> = {};
@@ -216,6 +218,7 @@ class Archive {
       const name = new TextDecoder('utf-8').decode(nameBytes);
       this.alloc(['class', name, schema]);
       this.classSlot.set(name, this.nextSlot - 1);
+      this.classSchema.set(name, schema);
       return this.newObj(r, name);
     }
     if (tag & 0x8000) {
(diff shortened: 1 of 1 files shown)


iamahsanmehmood:
**Review:** matches #41's (.NET) approach field-for-field — `classSchema`/`currentClass` tracking correctly threaded through `newObj`/`readInstance`. CI green (3/3). Verified against both #38 repro files plus 9 real production files (280–592 MB excluded — pre-existing Node/V8 heap limit unrelated to this change), zero regressions. Merging with admin privileges since GitHub blocks self-approval on same-account PRs.

PR #46, opened 2026-08-08
fix(dart): recognise big-tag CThumbnail class-refs in the definition-tail scan (ports #28)

iamahsanmehmood:
Ports the fix from #39 (Python) / #42 (.NET) / #45 (TypeScript) to the Dart legacy walker. Same root cause, same fix, different language.

## The bug

```dart
if (thumbSlot != null && Tlv.readU16(r.data, p) == (0x8000 | thumbSlot)) { ... }
```

Once the `CThumbnail` class slot passes `0x7FFF`, `0x8000 | slot` cannot fit in 16 bits, the comparison is dead code, and every definition tail throws `"thumbnail not found"` — while `readObject` had decoded the big-tag escape (`0x7FFF` + u32 of `0x80000000 | slot`) all along.

`CThumbnail`'s slot is the raw global object-slot number where its class was first registered — not a count of distinct class *types* — so any file where the first `CThumbnail` happens to be encountered after ~32,767 other objects have already been allocated hits this, independent of file size.

## The fix

Both encodings now live in one helper, `isClassRef`, mirroring `readObject` exactly (matches Python's `_is_class_ref` in #39). Unit tests cover the short form, the escape form, the truncated-data edge, and the specific pre-fix failure (a big slot must never match the truncated 16-bit comparison).

## Verification

- Swept 9 of the same 11 real production files used for prior regression testing (1.2 MB–109 MB): all parse identically before/after, zero regressions. The 2 #38 repro files correctly still fail with their original symptoms in this isolated build (they need the separate #38/#40 schema-gate fix, not this one).
- Full `dart test` suite: 19 passed (15 existing + 4 new).

Credit to @thomazyujibaba for the original TS report/repro (#28) and @tuxiasumari for confirming and porting to Python (#39).

### DIFF
--- a/packages/dart/lib/src/legacy.dart
+++ b/packages/dart/lib/src/legacy.dart
@@ -78,6 +78,19 @@ bool _matchesAscii(Uint8List data, int offset, String str) {
   return true;
 }
 
+/// True when the bytes at [p] are an MFC class-ref to class [slot]. Mirrors
+/// both encodings Archive.readObject decodes: the short 16-bit form
+/// (0x8000|slot) and, for slots past 0x7FFF, the big-tag escape (0x7FFF
+/// followed by a u32 of 0x80000000|slot).
+bool isClassRef(Uint8List data, int p, int slot) {
+  if (slot <= 0x7FFF) {
+    return p + 2 <= data.length && Tlv.readU16(data, p) == (0x8000 | slot);
+  }
+  return p + 6 <= data.length &&
+      Tlv.readU16(data, p) == 0x7FFF &&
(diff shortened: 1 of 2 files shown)


iamahsanmehmood:
**Review:** `isClassRef` correctly mirrors both encodings, matches #39/#42/#45. Test coverage includes the truncated-short-form regression case. No Dart CI exists in this repo; verified locally via `dart analyze`/`dart test` (19/19 passing, 4 new) plus zero regressions on the real-file corpus. Merging with admin privileges since GitHub blocks self-approval on same-account PRs.

PR #49, opened 2026-08-08
fix(cpp): SketchUp 2013 instances have no trailing GUID

thomasloockx:
## Summary
Fix the C++ legacy walker for the SketchUp 2013 instance layout identified in #38 and fixed for Python in #40.

## Type
- [x] Bug fix
- [ ] New feature
- [ ] Documentation

## Changes
- retain each MFC class schema while walking the legacy archive
- read the trailing GUID only for `CComponentInstance` schema 5+ and `CGroup` schema 1+
- preserve the existing GUID read for unknown schemas
- add regression coverage for both schema boundaries and unknown-schema behavior

## Validation
Both files attached to #38 now parse with finite geometry and match the Python fix results:

- `TestSketchup2013.skp`: 3 definitions, 69 faces, 1,617 vertices
- `TestSketchup8.skp`: 15 definitions, 39,274 faces, 26,286 vertices

Local CI-equivalent checks:

- GCC Release build: 44/44 tests passed
- Clang Release build: 44/44 tests passed
- install and downstream consumer builds passed with GCC and Clang
- ASan + UBSan Debug build: 44/44 tests passed
- `openskp-format-check` passed
- `git diff --check` passed

## Checklist
- [x] Code follows the project style guidelines
- [x] Tests pass locally
- [x] Documentation updated (not applicable)

### DIFF
--- a/packages/cpp/src/internal.hpp
+++ b/packages/cpp/src/internal.hpp
@@ -105,6 +105,7 @@ std::vector<std::pair<std::string, ByteBuffer>> parse_flat(const ByteBuffer&);
 std::string extract_version(const ByteBuffer&);
 bool valid_header(const ByteBuffer&);
 bool is_legacy(const ByteBuffer&);
+bool legacy_instance_has_guid(const std::string&, std::optional<int>);
 RawParsed full_parse(const ByteBuffer&, const ParseOptions&);
 RawParsed parse_legacy(const ByteBuffer&, const ParseOptions&);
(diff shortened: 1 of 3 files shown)


iamahsanmehmood:
Resolved the conflict against main (your version replaces the inline check I'd merged in #47 — yours correctly distinguishes 'schema unknown' from 'schema is genuinely 0' via std::optional, which mine didn't handle right). All CI green. Merging — thanks for catching this properly.

PR #53, opened 2026-08-08
fix(python): don't discard the legacy MFC material alpha byte

iamahsanmehmood:
`Material.color`'s alpha channel was silently dropped when parsing legacy MFC (SketchUp 2013–2020) files: `legacy.py` read the material's real 4-byte RGBA record (`rgba[0..3]`) but the dict it built only kept r/g/b, and `model.py`'s `Material(...)` construction only ever passed 3 values into a field typed `Tuple[int, int, int, int]` — a genuine type mismatch that silently produced a 3-tuple at runtime.

Verified empirically across 2,060 materials in 13 real production files (SketchUp 2013–2025) that this byte is always 255 in practice, but reading the real value is more correct/future-proof than assuming a constant — matches what the .NET port already does correctly.

The VFF (2021+) material XML record has no alpha attribute at all, so that path correctly continues to default to 255 (`c.get("a", 255)`).

One existing test (`test_face_material_resolves_through_materials_by_id`) was asserting the old, incorrect 3-tuple shape — fixed to expect the type-correct 4-tuple.

All 84 tests pass.

### DIFF
--- a/packages/python/src/openskp/legacy.py
+++ b/packages/python/src/openskp/legacy.py
@@ -870,7 +870,7 @@ def full_parse_legacy(skp_path: str) -> Dict[str, Any]:
         colorized = v.get('colorized', False)
         mat_obj: Dict[str, Any] = {
             'name': v['name'],
-            'color': {'r': rgba[0], 'g': rgba[1], 'b': rgba[2]},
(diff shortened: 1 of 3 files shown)


iamahsanmehmood:
Verified: all 84 tests pass, CI green across all Python versions/platforms.

PR #60, opened 2026-08-08
feat(python): bootstrap the legacy slot base for single- and zero-material files

tuxiasumari:
The legacy walker could only learn the absolute MFC slot base from files with **two or more materials** — material 2's class-ref names `CMaterial`'s true slot. Every other file raised `single-material bootstrap not implemented` (or `no CMaterial class record found` when the file has no materials at all). That is not an exotic case: it covers every template-fresh SketchUp drawing, and 37 of the 62 files in the independent Rust implementation's corpus fail on it today.

## The idea

The file itself names the base a few objects later. Right after the layer list comes the definition-list anchor: an **absolute** back-ref to the active layer — an object the walker just allocated **relatively**. So:

1. Parse the model prefix (materials, layer list) under a throwaway base.
2. Read the anchor's absolute tag `T`.
3. Each walked layer at relative slot `R` yields the candidate base `T − (R − B0)`; with a single layer (the common case) the answer is exact, and with several the full walk is simply retried per candidate.

Zero-material files additionally have no `CMaterial` record to anchor the walk on, so it now falls back to the first `CLayer` class record and starts at the layer-list marker just before it.

The two-material path is untouched and still preferred — and it doubles as the correctness oracle: on the 24 multi-material corpus files, the probe agrees with it **24 / 24**.

## Verification

- All 37 previously-failing corpus files parse, and the corpus names are the spec: `blank-template-box` → exactly 6 faces / 12 edges / 8 vertices; `curve` → 171 edges, 0 faces; `layers` → its 4 layers, through the multi-candidate path; `nested-3-deep` → 4 definitions.
- Wireframe renders of `box`, `cylinder`, `face-with-hole`, `nested-3-deep`, `box-v2013`, `box-v2016` all draw their names.
- The 25 previously-parsing legacy files produce **byte-identical vertex fingerprints** (order-stable SHA-256 over every coordinate) before and after.
- `pytest`: 80 passed on this branch.

Same gap presumably exists in the TS/.NET/Dart ports of the walker; the probe translates directly.

### DIFF
--- a/packages/python/src/openskp/legacy.py
+++ b/packages/python/src/openskp/legacy.py
@@ -647,27 +647,49 @@ def _walk(data: bytes):
     if not ver_m:
         raise LegacyParseError("no version string in header")
     ver = int(ver_m.group(1))
-    ar = _Archive(data, ver)
-    ar.readers.update(_READERS)
-    r = ar.r
-
     # anchor: the material manager (u32 count right before the first
-    # CMaterial new-class record)
+    # CMaterial new-class record); zero-material files have no CMaterial
+    # record anywhere, so fall back to the first CLayer class record and
+    # start at the layer-list marker just before it
     m = re.search(re.escape(b'\xff\xff') + b'..'
                   + re.escape(struct.pack('<H', 9) + b'CMaterial'),
                   data, re.DOTALL)
-    if not m:
-        raise LegacyParseError("no CMaterial class record found")
-    mat_hdr = m.start()
-    mat_count = struct.unpack_from('<I', data, mat_hdr - 4)[0]
-    if mat_count > 100000:
-        raise LegacyParseError("implausible material count")
+    if m:
+        start = m.start()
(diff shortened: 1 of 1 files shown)


iamahsanmehmood:
Reviewed thoroughly: the two-material bootstrap path is unchanged (just refactored into a named function, behaviorally identical), so zero regression risk there. Verified the new single/zero-material fallback logic by hand — the math checks out (each candidate base is solved from `anchor_absolute - (layer_relative - throwaway_base)`, with implausible-value bounds filtering, and the multi-candidate retry loop for ambiguous cases). Ran the full existing test suite against the actual merged result (not just this branch in isolation) — 85 passed. Also independently swept our own 13-file real-world corpus (1.2 MB–591.7 MB) against the merged state: zero regressions, all definitions/materials/layers counts identical to every prior run this session. The one failure (a 591.7 MB file hitting a MemoryError) is a pre-existing, unrelated resource limit — that file has many materials so it takes the untouched old bootstrap path entirely. CI green (13/13). Merging.

Issue #62, opened 2026-08-09
No uv coordinates exposed


japajoe:
Why does this class does not have any uv coordinates, or how am I supposed to get them when creating a scene?

```csharp
public sealed class GlbPrimitive
{
    /// <summary>Flat [x, y, z, x, y, z, ...] vertex positions, in
    /// metres, Y-up.</summary>
    public float[] Positions { get; set; } = Array.Empty<float>();

    /// <summary>Flat [x, y, z, ...] vertex normals, matching Positions
    /// 1:1.</summary>
    public float[] Normals { get; set; } = Array.Empty<float>();

    /// <summary>Triangle vertex indices into Positions/Normals (3 per
    /// triangle).</summary>
    public uint[] Indices { get; set; } = Array.Empty<uint>();

    /// <summary>Index into Scene.GltfMaterials for this primitive's
    /// resolved color.</summary>
    public int MaterialIndex { get; set; }

    /// <summary>Matches the corresponding key in Scene.MeshIndex.</summary>
    public string GeomName { get; set; } = "";
}
```

iamahsanmehmood:
The root cause: `GlbPrimitive` never computed texture coordinates at all — `Face.uvTransform` (the positioning data) was already decoded, but nothing in any language's scene-baking pipeline used it. This turned out to affect every language, not just C#.

Fixed and merged across all five ports (#63 Python, #64 .NET, #65 TypeScript, #66 C++, #67 Dart) — `GlbPrimitive` now has a `uvs`/`Uvs` field with real per-vertex UV coordinates, verified numerically identical across all five on the same real file. C++ additionally writes them into the actual exported `.glb` file's `TEXCOORD_0` accessor.

One known limitation, documented on the field: PROJECTED textures (terrain-drape, e.g. Add Location) still use the standard face-plane formula since the real projection-plane basis isn't captured in the parsed data yet — those UVs will be approximate.

PR #63, opened 2026-08-10
Python: compute real per-vertex UV coordinates in scene.GlbPrimitive

iamahsanmehmood:
## What

`scene.GlbPrimitive` exposed `positions`/`normals`/`indices` but no texture coordinates. `Face.uv_transform` (the positioning data) was already parsed and documented, but nothing in the scene-baking pipeline used it to actually compute UVs for the triangulated output.

Related to #62 (filed against C#'s `GlbPrimitive`, which has the same gap). This PR fixes Python only; the same fix is being ported to the other languages that expose this data shape separately, so the issue stays open until all of them land.

## How

- Added `GlbPrimitive.uvs`, computed per the formula already documented on `Face.uv_transform`: build a face-plane basis from the face normal, invert `uv_transform` when the face has a positioned/photo-fit mapping (default projection otherwise), divide by the material's tile size.
- Vertex dedup changed from keying on bare `v_id` to `(v_id, u, v)`. UV is inherently per-face - two faces sharing a vertex position can legitimately disagree on texture mapping there - and indexed glTF meshes require position/normal/uv aligned per vertex, so a vertex has to split wherever that happens.
- Merged the old, separate normals-accumulation pass (which re-walked every face's loops a second time) into the same loop that now computes UVs - a simplification that fell out of the restructuring, not separate scope.

## Known limitation

`uv_projected` faces (terrain-drape textures, e.g. Add Location) still use the face-plane formula, since the real projection-plane basis isn't captured anywhere in the parsed data - only a boolean flag is. Their UVs will be approximate. Documented in the field's docstring.

## Testing

- Full test suite: 85/85 passing, including the test that cross-validates mesh/material counts against the TypeScript reference on a real fixture (`capilla_quiroz_v17.skp`) - unchanged, as expected since UV doesn't affect face-color grouping.
- Ad-hoc verification against that same real fixture: no NaN/Inf across 782 output vertices, both the positioned-texture path (32 faces) and default-projection path (149 faces) exercised, neighboring vertices show coherent (not random) UV values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Python**: `scene.GlbPrimitive` gained a `uvs` field with real per-vertex
+  texture coordinates, computed from each source face's `uv_transform` (or
+  the default face-plane projection when a face has none) — see
+  `Face.uv_transform`'s docstring for the formula. Vertices are now split
+  where two faces sharing a position disagree on UV, since indexed glTF
+  meshes need position/normal/uv aligned per vertex. Fixes #62 for Python;
+  other languages exposing the same `GlbPrimitive` shape are being ported
+  separately. Faces with `uv_projected` set (terrain-drape textures) still
+  use the face-plane formula, since the real projection-plane basis isn't
+  captured anywhere in the parsed data yet — their UVs are approximate.
+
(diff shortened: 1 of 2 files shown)


iamahsanmehmood:
Verified: full test suite green across all OS/Python-version combinations, plus cross-checked the actual computed UV values against a real fixture — and against this same PR's C# port on the identical file (numerically identical UV coordinates, confirming the formula is implemented consistently). Merging.

PR #64, opened 2026-08-10
.NET: compute real per-vertex UV coordinates in Scene.GlbPrimitive

iamahsanmehmood:
## What

Related to #62 - this is the exact `GlbPrimitive` class the issue was filed against. It exposed `Positions`/`Normals`/`Indices` but no texture coordinates, so a consumer baking a scene had no way to actually texture the mesh even though `Face.UvTransform` (the positioning data) was already decoded and available.

Same root cause and same fix as #63 (already merged, Python). Other languages exposing this same `GlbPrimitive`/`Scene` shape (TypeScript, C++) are being ported separately in follow-up PRs - the issue stays open until all of them land.

## How

- Added `GlbPrimitive.Uvs`, computed per the formula documented on `Face.UvTransform`: build a face-plane basis from the face normal, invert `UvTransform` when the face has a positioned/photo-fit mapping (default projection otherwise), divide by the material's tile size.
- Vertex dedup changed from keying on bare `vId` to `(vId, u, v)`. UV is inherently per-face - two faces sharing a vertex position can legitimately disagree on texture mapping there - and indexed glTF meshes require position/normal/uv aligned per vertex, so a vertex has to split wherever that happens.
- Merged the old, separate normals-accumulation pass into the main triangulation loop, matching the Python restructuring - it was redundantly re-walking every face's loops a second time.

## Known limitation

`UvProjected` faces (terrain-drape textures) still use the face-plane formula, since the real projection-plane basis isn't captured anywhere in the parsed data - documented on the field.

## Testing

- `dotnet test`: 20/20 passing, including the existing real-fixture test that cross-validates mesh/material counts against Python/TypeScript (unchanged, as expected).
- Extended `PrimitivesHaveValidGeometry` to assert `Uvs` has the right length and no NaN/Inf, on the same real fixture.
- Strongest check: wrote a throwaway console app referencing this build, dumped the actual computed UV values on the same real fixture (`capilla_quiroz_v17.skp`) Python's PR used, and compared - **numerically identical UV values** (to float precision) between the two independent implementations, just in different vertex order (harmless - hash-map iteration order differs between languages).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   separately. Faces with `uv_projected` set (terrain-drape textures) still
   use the face-plane formula, since the real projection-plane basis isn't
   captured anywhere in the parsed data yet — their UVs are approximate.
+- **.NET**: `Scene.GlbPrimitive` gained a `Uvs` field, same fix as the
+  Python entry above and the exact issue #62 was filed against (its
+  `GlbPrimitive` snippet). Verified numerically identical to Python's
+  output (to float precision) on the same real fixture.
 
 - **C++**: public `to_glb(const Scene&)` and `export_glb(const Scene&, path)`
   APIs for in-memory and file-based binary glTF 2.0 export. The implementation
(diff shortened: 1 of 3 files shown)


iamahsanmehmood:
Verified: full 3-OS test suite green, plus cross-checked the actual computed UV values against Python's PR #63 on the identical real fixture — numerically identical (to float precision). Merging.

PR #65, opened 2026-08-10
TypeScript: compute real per-vertex UV coordinates in GlbPrimitive

iamahsanmehmood:
## What

Related to #62 - same gap as #63 (Python) and #64 (.NET, the exact class the issue was filed against), now ported to TypeScript. `GlbPrimitive` exposed `positions`/`normals`/`indices` but no texture coordinates, so a consumer baking a scene had no way to actually texture the mesh even though `Face.uvTransform` (the positioning data) was already decoded and available.

C++ is being ported separately - the issue stays open until it lands too.

## How

Same approach as the two merged PRs:

- Added `GlbPrimitive.uvs`, computed per the formula documented on `Face.uvTransform`: face-plane basis from the normal, invert `uvTransform` when present, divide by material tile size.
- Vertex dedup changed from bare `vId` to `(vId, u, v)` - UV is inherently per-face, so a shared vertex position has to split wherever two faces disagree on its texture mapping.
- Merged the old, separate normals-accumulation pass into the main triangulation loop, matching the Python/.NET restructuring.

## Known limitation

PROJECTED textures (terrain-drape, e.g. Add Location) still use the face-plane formula, since the real projection-plane basis isn't captured anywhere in the parsed data - documented on the field, matching the other ports.

## Testing

- `npm test`: 40/40 passing (added a new test asserting every baked primitive's `uvs` has the right length and no NaN/Inf on the real fixture).
- `tsc --noEmit` and `npm run build` both clean.
- Strongest check: dumped the actual computed UV values on the same real fixture (`capilla_quiroz_v17.skp`) via a throwaway test, compared against Python's and .NET's already-merged output on the identical file - **numerically identical UV values** (to float precision) across all three independent implementations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Python entry above and the exact issue #62 was filed against (its
   `GlbPrimitive` snippet). Verified numerically identical to Python's
   output (to float precision) on the same real fixture.
+- **TypeScript**: `GlbPrimitive` gained a `uvs` field, same fix as the
+  Python/.NET entries above. Verified numerically identical to both on the
+  same real fixture.
 
 - **C++**: public `to_glb(const Scene&)` and `export_glb(const Scene&, path)`
   APIs for in-memory and file-based binary glTF 2.0 export. The implementation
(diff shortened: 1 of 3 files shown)


iamahsanmehmood:
Verified: full test suite green on Node 18/20/22, plus cross-checked the actual computed UV values against Python's (#63) and .NET's (#64) merged output on the identical real fixture — numerically identical (to float precision) across all three. Merging.

PR #66, opened 2026-08-10
C++: compute real per-vertex UV coordinates in GlbPrimitive and GLB output

iamahsanmehmood:
## What

Related to #62 - same gap as #63 (Python), #64 (.NET), #65 (TypeScript), now ported to C++. `GlbPrimitive` exposed `positions`/`normals`/`indices` but no texture coordinates.

C++ is the last of the four languages that already ship GLB export - this closes the loop for #62. (Dart/.NET-file-writer scope is separate/pre-existing, tracked independently.)

## How

Same formula as the other three merged PRs (face-plane basis from the normal, invert `uv_transform` when present, divide by material tile size), with two C++-specific differences from how this repo's C++ port is structured:

1. **`glb.cpp` wiring.** Unlike Python (whose real `.glb` writer is a separate legacy trimesh path not addressed here) and TS/.NET (whose `GlbPrimitive` is consumed by callers, not an internal file writer), C++'s `to_glb`/`export_glb` write real `.glb` files *directly* from `GlbPrimitive`. So this PR also wires a `TEXCOORD_0` accessor into the actual glTF output in `glb.cpp` - exported `.glb` files now carry real texture coordinates, not just the in-memory struct.

2. **Front/back materials.** C++ uniquely models a face's front and back materials as two separate primitives when they differ (`add_side` with a reverse-winding flag) - none of the other four ports do this at all. The back-side primitive, when present, computes UV from `uv_transform_back` using the same face-plane basis as the front, per the documented recipe ("same for the back side"). I want to flag honestly: **this specific path has no cross-language reference to verify against**, since no other port models back materials. It's a literal reading of the documented formula, not an invented heuristic, but it's the one part of this PR I couldn't cross-check numerically against an already-merged implementation the way I could for the front-facing/single-sided case.

Vertex dedup key changed from vertex id alone to `(vertex id, u, v)` - UV is inherently per-face, so a shared vertex has to split wherever two faces disagree on its mapping.

## Verification (please weigh this carefully - see below)

**I cannot compile or run this locally** - there is no C++ toolchain (no cmake, no compiler) in this environment, same constraint as every prior C++ PR in this repo. Given that, and the added complexity here (vertex-key restructuring, a nested lambda, front/back material handling), I did more manual work than usual before opening this:

- Ran `clang-format` locally (pip-installed 22.1.8, vs. CI's clang-format-18) against every changed file - clean except one whitespace diff in a wrapped lambda signature, now fixed.
- Caught and fixed a real C++17 pitfall myself: the face-plane basis (`xr`, `yr`) is computed once per face and then used inside a nested `add_side` lambda - originally written as a structured binding (`const auto [xr, yr] = ...`), which cannot be captured by a lambda before C++20. Rewrote as plain named references.
- Line-by-line compared the vertex-dedup and UV-computation logic against the already-merged, already-cross-verified Python/.NET/TypeScript implementations for structural parity.
- Extended the existing `glb_test.cpp` and `parser_test.cpp` coverage (uvs length/finiteness checks, `TEXCOORD_0` accessor assertions, malformed/non-finite-uv rejection tests, matching the existing test shape for positions/normals) rather than adding a separate, disconnected test file.

I'm relying on CI's 3-OS build+test matrix as the actual compile/correctness gate here, same as prior C++ work - flagging this explicitly rather than overstating confidence.

## Testing

- `clang-format --dry-run --Werror` equivalent: clean locally (see above).
- Test file updates: `glb_test.cpp`'s `triangle_scene()` helper uses aggregate initialization, so adding `uvs` in the middle of the struct required updating it with a placeholder UV array - done, and extended with `TEXCOORD_0` assertions plus malformed/non-finite rejection cases mirroring the existing `positions`/`normals` tests. `parser_test.cpp`'s real-fixture loop now also asserts `uvs` length and finiteness.
- Will confirm CI is fully green before merging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **TypeScript**: `GlbPrimitive` gained a `uvs` field, same fix as the
   Python/.NET entries above. Verified numerically identical to both on the
   same real fixture.
+- **C++**: `GlbPrimitive` gained a `uvs` field, same fix as the other three
+  ports. Unlike those, C++'s `to_glb`/`export_glb` write real `.glb` files
+  directly from this struct, so this PR also wires a `TEXCOORD_0` accessor
+  into the actual glTF output - exported files now carry real texture
+  coordinates, not just the in-memory struct. C++ also uniquely models
+  front/back materials as separate primitives; the back-side primitive
+  (when present) uses `uv_transform_back` with the same face-plane basis
+  as the front, per the documented recipe - this path has no cross-language
+  reference to verify against, since no other port models back materials
+  at all.
 
 - **C++**: public `to_glb(const Scene&)` and `export_glb(const Scene&, path)`
   APIs for in-memory and file-based binary glTF 2.0 export. The implementation
--- a/packages/cpp/include/openskp/scene.hpp
+++ b/packages/cpp/include/openskp/scene.hpp
@@ -31,6 +31,17 @@ struct MeshMetadata {
 struct GlbPrimitive {
   std::vector<float> positions;
   std::vector<float> normals;
+  // Flat [u, v, u, v, ...] texture coordinates, matching positions 1:1.
+  // Computed from the source face's uv_transform (or the default
+  // face-plane projection when a face has none): a face-plane basis from
+  // the normal, inverting uv_transform when present, divided by the
+  // material's texture tile size. A vertex shared by two faces that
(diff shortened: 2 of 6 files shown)


iamahsanmehmood:
All CI green: GCC, Clang (Linux+macOS), MSVC, ASan/UBSan sanitizers, and clang-format — across both the format-fix and the follow-up push. This validates the two things I flagged as unverifiable locally (the C++17 structured-binding-capture fix, and the new tuple-keyed vertex dedup) actually compile and run clean on every supported compiler. Merging.

PR #67, opened 2026-08-10
Dart: compute real per-vertex UV coordinates in GlbPrimitive

iamahsanmehmood:
## What

Related to #62 - same gap as #63 (Python), #64 (.NET), #65 (TypeScript), #66 (C++), now ported to Dart, the last language whose scene-baking layer exposes `GlbPrimitive`. `GlbPrimitive` exposed `positions`/`normals`/`indices` but no texture coordinates.

This is the last of the five - once this merges, #62 is resolved everywhere the data shape it's about actually exists.

## How

Same formula as the four merged PRs: face-plane basis from the normal, invert `uvTransform` when present, divide by material tile size. Vertex dedup key changed from vertex id alone to `(vertex id, u, v)` - Dart 3 records are structurally hashable so they work directly as `Map` keys, no extra plumbing needed. Merged the old, separate normals-accumulation pass into the main triangulation loop, matching the other ports.

Dart has no `.glb` file writer yet (that's tracked separately, alongside .NET, as its own follow-up task) - this reaches the `GlbPrimitive` data shape only, same scope as the Python/.NET/TypeScript PRs.

## Testing

- `dart test`: all 6 test files pass individually (a local quirk means bare `dart test` in this environment only auto-discovers one file - ran each explicitly to get full coverage; CI's `ci-dart.yml` invocation should not have this problem). Extended `scene_test.dart`'s existing real-fixture geometry test with `uvs` length/finiteness assertions, matching the pattern already used for `positions`/`normals`.
- Strongest check: dumped the actual computed UV values on the same real fixture (`capilla_quiroz_v17.skp`) via a throwaway script, compared against Python's/.NET's/TypeScript's/C++'s already-merged output on the identical file - **numerically identical UV values** (to float precision), matching even the exact vertex ordering C# produced.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   as the front, per the documented recipe - this path has no cross-language
   reference to verify against, since no other port models back materials
   at all.
+- **Dart**: `GlbPrimitive` gained a `uvs` field, same fix as the other
+  ports. This closes out #62 across every language whose scene-baking
+  layer exposes `GlbPrimitive` - Python, .NET, TypeScript, C++, and now
+  Dart all compute numerically identical UV values (to float precision) on
+  the same real fixture. Dart has no `.glb` file writer yet (tracked
+  separately), so this reaches the `GlbPrimitive` data shape only, same as
+  the Python/.NET/TypeScript entries above.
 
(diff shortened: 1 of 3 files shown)


iamahsanmehmood:
Verified: full 3-OS test suite green (also confirms CI's dart test invocation runs the whole suite fine, unlike a local quirk I hit). Cross-checked computed UV values against Python/.NET/TypeScript/C++'s merged output on the identical real fixture — numerically identical (to float precision), same ordering as C#. Merging — this closes out the language ports for #62.

PR #68, opened 2026-08-10
Python: real UV coordinates + fix crash in export.glb.export()'s writer

iamahsanmehmood:
## What

Follow-up to #62 (Python's part already merged in #63, fixing the `scene.GlbPrimitive` data shape). This PR fixes the *other* Python GLB pipeline: `export.glb.export()` - the actual `.glb` file writer, a separate legacy trimesh-based implementation, not `scene.py`'s `build_scene()`. It had two real, previously-undetected bugs, found while verifying it end to end for the first time (there was no existing test coverage for this function at all).

## Bug 1: crash on any textured file

The metadata JSON sidecar this function writes tried to embed each material's raw texture image bytes directly - never valid JSON. `export()` crashed with `TypeError` on any real file with a textured material. Fixed by stripping the binary `data` field before serialization (the `.glb` file itself carries the actual texture bytes; a JSON metadata file was never the right place for them). Everything else in the metadata schema - `mesh_index`, `scene_hierarchy` (including `definition_id`/`matrix_3x4`), `layers` - is untouched.

## Bug 2: no UV coordinates at all, and a real color-fidelity constraint underneath it

Every face of a definition went into a single trimesh mesh with a flat *per-face* color array, regardless of how many distinct materials were mixed into that definition - confirmed on the real fixture, 2 of 3 definitions mix colors. trimesh's UV/texture visual (`TextureVisuals`) carries one material per mesh, so before UV could be added at all, faces needed to be grouped into one mesh per resolved color per definition - the same grouping `scene.py`'s `build_scene()` already uses. UV itself is computed via the identical face-plane formula (reusing `scene.py`'s `_face_uv_basis`/`_compute_face_uv` via a local import inside the function, to avoid a circular import between `_core.py` and `scene.py`).

One more thing this surfaced: trimesh's `Trimesh` constructor does automatic post-construction vertex processing (merging/dropping vertices by position) unless you pass `process=False`. Since the UV computation splits vertices by `(vertex, uv)` rather than position alone, that default processing was silently misaligning - and for several meshes, entirely dropping - the UV array trimesh's glTF exporter has a `len(uv) == len(vertices)` check and *silently* omits `TEXCOORD_0` when it fails, no error or warning. Caught this by actually inspecting the raw glTF JSON in the exported file, not just checking "did it run without an exception."

## Testing

- Full test suite: 87/87 (added 2 new tests for `export.glb.export()` - previously zero coverage).
- Verified end to end on three real files (two legacy MFC, one modern VFF format, up to 43 meshes / ~14,800 vertices): every exported mesh has real UV data (checked the raw glTF JSON directly, not just a reloaded/round-tripped object), mesh counts and computed UV values match `scene.py`'s already-verified `build_scene()` output exactly on the same files, and the metadata JSON's existing schema is unchanged (`definition_id`/`matrix_3x4` still present).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the same real fixture. Dart has no `.glb` file writer yet (tracked
   separately), so this reaches the `GlbPrimitive` data shape only, same as
   the Python/.NET/TypeScript entries above.
+- **Python**: `export.glb.export()` - the actual `.glb` file writer, a
+  separate legacy pipeline from `scene.py`'s `build_scene()` - now writes
+  real per-vertex UV coordinates too, closing the same gap as above for
+  Python's literal exported files (matching what C++'s PR already did for
+  its own file writer). Faces are now grouped into one mesh per resolved
+  color per definition, same as `scene.py`, since a single trimesh mesh
+  can only carry one material - previously this pipeline put every face
+  of a definition into one mesh with a flat per-face color array,
+  regardless of how many distinct materials were mixed in (confirmed on a
+  real fixture: 2 of 3 definitions mix colors). Verified numerically
+  identical UV output to `scene.py`'s on three real files (both legacy
+  MFC and modern VFF format).
 
 - **C++**: public `to_glb(const Scene&)` and `export_glb(const Scene&, path)`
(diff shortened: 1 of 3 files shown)


iamahsanmehmood:
Verified: full test suite green across all OS/Python-version combinations. This closes the loop on the actual exported .glb file for Python, matching what C++'s TEXCOORD_0 wiring already did for its own writer — real UV data now reaches literal output files in both languages that have one. Merging.

PR #69, opened 2026-08-10
TypeScript: expose Face.uvProjected/uvProjectedBack

iamahsanmehmood:
## What

`Face` gained `uvProjected`/`uvProjectedBack`. The legacy MFC reader already decoded these bits internally (`front_projected`/`back_projected` on the `CFaceTextureCoords` record - used for PROJECTED textures like the Add Location terrain drape) but discarded them instead of exposing them on the public model. Same gap Python closed for itself in #61.

A PROJECTED texture has UVs that run in the projection plane's frame rather than the standard face frame, so a renderer needs to know about it to map the texture correctly - right now there's no way to tell.

VFF/modern files don't carry this flag at all in any language so far - defaults to `false` there, matching Python's precedent.

## Testing

- `tsc --noEmit`: clean.
- `npm test`: 40/40 passing.
- Honest disclosure: neither this fix nor Python's original #61 has a dedicated unit test for the true/projected case - the real test fixture has zero projected faces (confirmed by checking Python's parsed output), and hand-crafting a synthetic byte-accurate `CFaceTextureCoords` record for a unit test is nontrivial. Matched the existing bar rather than silently skipping the gap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **TypeScript**: `Face` gained `uvProjected`/`uvProjectedBack` fields.
+  The legacy MFC reader already decoded these bits internally
+  (`front_projected`/`back_projected` on the `CFaceTextureCoords` record)
+  but discarded them instead of exposing them, same gap Python closed in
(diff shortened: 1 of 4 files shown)


iamahsanmehmood:
Verified: full test suite green on Node 18/20/22. Merging.

PR #70, opened 2026-08-10
Dart: expose Face.uvProjected/uvProjectedBack

iamahsanmehmood:
## What

Same gap as #69 (TypeScript, merged): `Face` gained `uvProjected`/`uvProjectedBack`. The legacy MFC reader already decoded these bits (`frontProjected`/`backProjected` on the `CFaceTextureCoords` record) but discarded them instead of exposing them - same gap Python closed for itself in #61.

VFF/modern files don't carry this flag - defaults to `false` there, matching Python's precedent.

## Testing

- `dart test`: all 6 test files pass individually.
- Same honest disclosure as #69: no dedicated unit test for the true/projected case in this PR or Python's original - the real fixture has zero projected faces, and hand-crafting a synthetic legacy record for one is nontrivial. Matching the existing bar rather than silently skipping the gap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   need to know this to render it correctly. VFF/modern files don't carry
   this flag at all, so it correctly defaults to `false` there, matching
   Python's precedent.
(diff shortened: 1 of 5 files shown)


iamahsanmehmood:
Verified: full 3-OS test suite green. Merging.

PR #71, opened 2026-08-10
.NET: expose Face.UvProjected/UvProjectedBack

iamahsanmehmood:
## What

Same gap as #69 (TypeScript) and #70 (Dart), both merged: `Face` gained `UvProjected`/`UvProjectedBack`. The legacy MFC reader already decoded these bits (`FrontProjected`/`BackProjected` on the `CFaceTextureCoords` record) but discarded them instead of exposing them - same gap Python closed for itself in #61.

VFF/modern files don't carry this flag - defaults to `false` there, matching Python's precedent. This is the last of the four remaining languages (C++ needs a separate, bigger fix - see follow-up).

## Testing

- `dotnet test`: 20/20 passing.
- Same honest disclosure as #69/#70: no dedicated unit test for the true/projected case - the real fixture has zero projected faces in any language checked so far.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Python's precedent.
 - **Dart**: `Face` gained `uvProjected`/`uvProjectedBack` fields, same fix
   and same rationale as the TypeScript entry above.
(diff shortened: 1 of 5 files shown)


iamahsanmehmood:
Verified: full 3-OS test suite green. Merging — this closes out TS/Dart/.NET for the projected-flag gap. C++ needs a separate, larger fix (report incoming).

PR #72, opened 2026-08-10
C++: fix legacy uv_transform never being populated; expose uv_projected

iamahsanmehmood:
## What

Started as a smaller task - expose the already-decoded PROJECTED-texture bit, matching TypeScript/Dart/.NET's `Face.uvProjected`/`UvProjected` (#69, #70, #71, all merged) - and turned into a real, larger correctness bug.

`CFace`'s attribute container - the MFC record that holds its `CFaceTextureCoords` positioning mapping, alongside any other attribute dictionaries - was being read (correctly advancing the archive cursor, so nothing crashed or looked corrupted) and then immediately discarded via `Archive::preamble()`, never linked back to the owning face. This means **`Face.uv_transform`/`uv_transform_back` have never been populated for any legacy MFC file in this port** - every legacy face's UV has silently used the default (non-positioned) face-plane projection, even when the SketchUp author explicitly positioned or photo-fitted a texture.

Nothing about this looked wrong from the outside - geometry, materials, and the *default*-projection UVs were all still correct - which is exactly why it went unnoticed, including through this session's own earlier #62 work (#66), which explicitly disclosed it couldn't verify C++'s UV output numerically against Python's (no local compiler available). This PR is the concrete bug that disclosure was flagging as a risk.

## How

- `preamble()` now returns the attribute container's slot (`std::optional<uint64_t>`, since slot 0 is a legitimate real slot - the first object allocated in the archive - so it can't double as an "absent" sentinel) instead of `void`.
- `CFace` captures this into `V::attrs` (a field that was already declared but dead - never written or read anywhere).
- `CAttributeContainer`'s own read loop now captures each child object into `V::ents` instead of discarding it too. `CFaceTextureCoords` can appear there directly - `Archive::object()`'s `expect` parameter is only a fallback used when defining a brand-new class slot the first time it's seen, not a hard type constraint on backrefs to already-known slots.
- `CFaceTextureCoords` now also captures its two trailing flag words (previously read-and-discarded) and decodes the PROJECTED bit from each.
- `fill()`'s face-building resolves the attrs slot, scans its children for one tagged `"ftc"`, and populates `uv_transform`/`uv_transform_back`/`uv_projected`/`uv_projected_back` from it - mirroring exactly how Python/TypeScript/Dart/C# already do this linkage.

`Face` gained `uv_projected`/`uv_projected_back`, matching the other three languages' fields from this same PR series. VFF/modern files don't carry this flag - defaults to `false` there.

## Verification (please weigh this carefully)

**Cannot compile or run this locally** - no C++ toolchain in this environment, same constraint as every prior C++ PR here. Given this touches parsing/linkage logic rather than adding a field on top of already-correct data (which is what my prior C++ PRs did), I did more than usual before opening this:

- `clang-format` applied and clean.
- Traced every changed line against Python's already-correct, already-shipped implementation of this exact linkage mechanism (`CAttributeContainer` → `CAttributeNamed`/`CFaceTextureCoords` children, resolved by slot) to confirm the resolution logic matches structurally, not just superficially.
- Added a real regression test: `Parser.LegacyMatchesReference` now asserts the total count of faces with a real `uv_transform` across the whole real fixture equals **32**, matching Python's independently-verified count on the identical file exactly. Before this fix, that count would always have been 0, for every legacy file, silently.

Relying on CI's full compiler matrix (GCC, Clang, MSVC, plus ASan/UBSan sanitizers) as the actual compile/correctness gate - will confirm fully green before merging, same as every prior C++ PR in this repo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and same rationale as the TypeScript entry above.
 - **.NET**: `Face` gained `UvProjected`/`UvProjectedBack` properties, same
   fix and same rationale as the TypeScript/Dart entries above.
+- **C++**: `Face` gained `uv_projected`/`uv_projected_back` fields, same
+  rationale as the TypeScript/Dart/.NET entries above. See the Fixed
+  section for why this needed more than exposing two bits.
 
 - **Python**: `scene.GlbPrimitive` gained a `uvs` field with real per-vertex
   texture coordinates, computed from each source face's `uv_transform` (or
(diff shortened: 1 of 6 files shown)


iamahsanmehmood:
All CI green: GCC, Clang (Linux+macOS), MSVC, ASan/UBSan sanitizers, format — across both pushes. The regression test I added is part of the standard suite that ran on every job, so this is real numeric confirmation: the fixture now shows exactly 32 faces with a real uv_transform, matching Python's independently-verified count on the identical file exactly (would have been 0 before this fix, for every legacy file, silently). Sanitizers passing is a good sign for the shared_ptr lifetime changes specifically. Merging.

PR #73, opened 2026-08-10
Python: add model.root to JSON export; fix 5 lint issues in tests

iamahsanmehmood:
## What

Two small, related cleanup items from the repo-health checklist:

1. **`export.json_export.to_dict()` was missing `root`.** Its output only included `definitions` (the numeric-ID-keyed component/group definitions), silently omitting any geometry/instances placed directly in the model rather than inside a component. `model.root` has been its own dataclass field since #59, but the JSON exporter was never updated to include it.

2. **5 pre-existing ruff issues in `test_parser.py`**, invisible to CI since `ci-python.yml`'s lint step only checks `src/`, not `tests/`: 4 unused imports (`typing.List`, and three local `pathlib`/`struct` imports genuinely never referenced in their function bodies - all the `pathlib.Path` usages elsewhere are type annotations, which `from __future__ import annotations` makes lazy strings that never need a runtime import) and one `import pathlib` statement misplaced mid-file instead of at the top (moved up rather than removed, since `pathlib` is genuinely used throughout the file's annotations - ruff correctly recognized it as used, just badly placed).

## Testing

- Full test suite: 88/88 passing (added a real test for the `root` JSON key, and extended the existing empty-model test to check it too).
- `ruff check src/ tests/`: fully clean now (previously only `src/` was clean).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Python**: `export.json_export.to_dict()`'s output gained a top-level
+  `root` key with the model's implicit top-level definition (matching
+  `SkpModel.root`) - previously only `definitions` (the numeric-ID-keyed
+  component/group definitions) was included, so the JSON export silently
+  omitted any geometry/instances placed directly in the model rather than
+  inside a component.
+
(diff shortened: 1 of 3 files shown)


iamahsanmehmood:
Verified: full test suite green across all OS/Python-version combinations, lint clean including tests/ now. Merging.

PR #74, opened 2026-08-10
.NET: add binary glTF (GLB) export - ToGlb/ExportGlb

iamahsanmehmood:
## What

Ports the last missing capability from Python/TypeScript/C++: a way to actually export a baked `Scene` as a `.glb` file, not just the intermediate `Scene`/`GlbPrimitive` data. Matches Python's and C++'s `to_glb`/`export_glb` pair (TypeScript only has the bytes-returning variant, no file-write one - so "full parity" here means matching the fullest existing API surface, not literally every other language, since they aren't uniform).

## How

A from-scratch writer, no new dependency - matching how this project has stayed dependency-light everywhere except C++'s bundled TinyGLTF. `netstandard2.0` has no built-in JSON support, so this also adds a small internal `MiniJson` serializer: reflection-based, only covering what this writer actually needs (Dictionary/List/array/primitive, plus anonymous types for `Scene.GltfMaterials` - anonymous-type property names match their declared identifiers exactly, so reflecting them produces correct camelCase glTF keys with no name mapping needed). Not a general-purpose JSON library, just enough for this.

Structurally ported from the TypeScript reference implementation's `toGLB()` (chunk framing, accessor/bufferView construction, min/max bounds on POSITION), with one real correction worth flagging: **TS's own `toGLB()` still doesn't write `TEXCOORD_0`**, despite `GlbPrimitive.uvs` existing on its data model too (added back in #65). I'm treating that as a separate, small follow-up rather than folding it into this PR. This new .NET writer includes `TEXCOORD_0` from the start, since building it fresh made that the obviously right thing to do.

`ExportGlb` deliberately does **not** create missing parent directories, matching C++'s `export_glb` - the only other language with this exact in-memory-bytes/file-write pair to stay consistent with. (Python's `export.glb.export()` does create directories, but it's a heavier, older pipeline that also writes JSON metadata/thumbnails alongside the GLB - not a direct analog to compare against here.)

## Verification

Built a throwaway console app referencing this build, exported the real fixture (`capilla_quiroz_v17.skp`), and inspected the raw GLB bytes directly (not just "did `ToGlb` return without throwing"):
- magic/version/chunk headers correct
- 13 meshes / 9 materials - matching every other language's already-verified count on this exact file
- all 13 primitives carry `TEXCOORD_0`
- the decoded UV values are **byte-for-byte identical** to the already cross-verified Python/TypeScript/C#/Dart output on the same file

## Testing

- `dotnet test`: 26/26 passing (20 existing + 6 new).
- New `GlbTests.cs` covers: serialization shape (POSITION/NORMAL/TEXCOORD_0 accessors, min/max bounds, material pass-through), empty scenes, malformed-geometry and non-finite-value rejection (mirroring C++'s `glb_test.cpp`'s coverage), and a real-fixture round-trip test that decodes every primitive's `TEXCOORD_0` back out of the binary chunk and asserts it matches the source `GlbPrimitive.Uvs` array exactly - not just "some accessor exists."
- Clean build, 0 warnings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **.NET**: `GlbExport.ToGlb(Scene)`/`GlbExport.ExportGlb(Scene, path)` -
+  binary glTF 2.0 (GLB) export, matching Python's and C++'s
+  `to_glb`/`export_glb` pair (TypeScript only has the bytes-returning
+  variant). A from-scratch writer with no new dependency, matching how
+  this project has stayed dependency-light everywhere except C++'s
+  bundled TinyGLTF - `netstandard2.0` has no built-in JSON support, so
+  this also adds a small internal `MiniJson` serializer (reflection-based,
+  just enough to cover the object graphs this writer builds; not a
+  general-purpose JSON library). Ported from the TypeScript reference
+  implementation's `toGLB()`, with one correction: TS's own `toGLB()`
+  still doesn't write `TEXCOORD_0` despite `GlbPrimitive.uvs` existing
+  there too (tracked as a separate follow-up) - .NET's writer includes it
+  from the start. `ExportGlb` doesn't create missing parent directories,
+  matching C++'s `export_glb` (the other language with this same pair).
+  Verified against a real fixture: magic/chunk headers, mesh/material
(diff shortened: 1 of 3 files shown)


iamahsanmehmood:
Verified: full 3-OS test suite green (26/26 including 6 new GLB tests), plus manual byte-level inspection of a real exported .glb file confirming correct chunk structure and UV values matching Python/TS/C#/Dart exactly. Merging.

PR #75, opened 2026-08-10
Dart: add binary glTF (GLB) export - toGlb/exportGlb

iamahsanmehmood:
## What

Same gap as #74 (.NET, merged): ports a way to actually export a baked `Scene` as a `.glb` file, not just the intermediate `Scene`/`GlbPrimitive` data. Matches Python's and C++'s `to_glb`/`export_glb` pair.

## How

A from-scratch writer, no new dependency. `dart:convert`'s built-in `jsonEncode` covers the JSON chunk directly - no custom serializer needed, simpler than the .NET port, which had to hand-roll one since `netstandard2.0` has no built-in JSON support.

Structurally ported from the TypeScript reference implementation's `toGLB()`, with the same `TEXCOORD_0` correction as the .NET port: TS's own `toGLB()` doesn't write it despite `GlbPrimitive.uvs` existing on its data model too (tracked as a separate follow-up, not part of this PR). This writer includes it from the start.

`exportGlb` doesn't create missing parent directories, matching C++/.NET's `export_glb`/`ExportGlb` - Dart's `File.writeAsBytesSync` already behaves this way by default, nothing extra needed.

One Dart-specific wrinkle worth flagging: `GlbPrimitive`'s `positions`/`normals`/`uvs` are `List<double>` (64-bit) here, unlike TS's `Float32Array` or C#'s `float[]` - but the binary accessor data is float32. Computing min/max bounds from the raw doubles would have produced values that don't match what's actually stored in the POSITION accessor's own binary data. Fixed by reading the bounds back from the already-written float32 buffer instead of the source doubles.

## Testing

- `dart test`: all 7 test files pass individually (same local-quirk workaround as prior Dart PRs - bare `dart test` only auto-discovers one file in this environment; CI's invocation doesn't have this problem, confirmed on #70/#71).
- New `glb_test.dart` (6 tests) mirrors the C++/.NET GLB test coverage: serialization shape, empty scenes, malformed-geometry and non-finite-value rejection, and a real-fixture round-trip test that decodes every primitive's `TEXCOORD_0` back out of the binary chunk and asserts it matches the source `GlbPrimitive.uvs` (accounting for float32 rounding, made explicit in the test itself).
- Verified end to end against the real fixture via a throwaway script: chunk headers correct, 13 meshes/9 materials matching every other language, `TEXCOORD_0` on all primitives, decoded UV values matching the already cross-verified Python/TypeScript/C#/.NET output on the same file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Dart**: `toGlb(Scene)`/`exportGlb(Scene, path)` - binary glTF 2.0
+  (GLB) export, matching Python's and C++'s `to_glb`/`export_glb` pair
+  (same scope as the .NET entry below). A from-scratch writer with no new
+  dependency - `dart:convert`'s built-in `jsonEncode` covers the JSON
+  chunk directly, no custom serializer needed (simpler than .NET's port,
+  which had to hand-roll one since `netstandard2.0` has no built-in JSON
+  support). Same TEXCOORD_0 correction as .NET's port relative to the
+  TypeScript reference. `exportGlb` doesn't create missing parent
+  directories, matching C++/.NET's `export_glb`/`ExportGlb`. One
+  Dart-specific wrinkle: `GlbPrimitive`'s fields are `List<double>`
+  (64-bit) here, but the binary accessor data is float32 - min/max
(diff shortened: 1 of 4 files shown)


iamahsanmehmood:
Verified: full 3-OS test suite green (all 7 test files pass), plus manual byte-level inspection of a real exported .glb file confirming correct chunk structure and UV values matching Python/TS/C#/.NET. Merging — this closes out GLB export across all 5 languages.

PR #76, opened 2026-08-10
TypeScript: write TEXCOORD_0 in toGLB()'s actual exported GLB bytes

iamahsanmehmood:
## What

`toGLB()` never wrote a `TEXCOORD_0` accessor, despite `GlbPrimitive.uvs` existing on the data model since #65 - real UV data was computed but the actual exported `.glb` bytes never carried it.

Found while using `toGLB()` as the structural reference for porting GLB export to .NET (#74) and Dart (#75) - both new writers included `TEXCOORD_0` from the start rather than reproducing this gap, and this is the follow-up to close it here too.

Same shape as the gap already fixed for Python's `export/glb.py` (#68) and C++'s `glb.cpp` (#66): the field existed on the data model, but the actual file/byte writer never read it.

## Testing

No test coverage existed for `toGLB()` before now, which is how this went unnoticed. Added real coverage in `glb.test.ts`:
- Serialization shape - accessors, min/max bounds, material pass-through, now including `TEXCOORD_0`.
- Empty-scene case.
- A real-fixture round-trip test that decodes every primitive's `TEXCOORD_0` back out of the binary chunk and confirms it matches the source `GlbPrimitive.uvs` exactly.

`tsc --noEmit` and `npm run build` both clean. Full suite: 43/43 passing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -123,6 +123,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **TypeScript**: `toGLB()` never wrote a `TEXCOORD_0` accessor, despite
+  `GlbPrimitive.uvs` existing on its data model since #65 - real UV data
+  was computed but the actual exported `.glb` bytes never carried it.
+  Found while using `toGLB()` as the structural reference for porting GLB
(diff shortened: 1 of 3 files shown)


iamahsanmehmood:
Verified: full test suite green on Node 18/20/22. Merging — TEXCOORD_0 is now written correctly across all 5 languages' GLB writers.

PR #77, opened 2026-08-10
docs: fix stale claims about Dart/.NET GLB export and other drift

iamahsanmehmood:
## What

Item 1 (and a few closely-related bonus fixes found while touching the same sections) from a full-repository audit conducted earlier today. This one was time-sensitive: `README.md` and `docs/DEVELOPER_GUIDE.md` said Dart and .NET "expose scene data only" / GLB export is "not yet ported" - both shipped complete, tested GLB writers earlier the same day. Actively wrong, not just stale.

## Changes

- **GLB export claims** - feature matrix, export-capabilities table, architecture diagram, and Quick Start snippets for .NET/Dart now correctly show GLB export. Verified the exact snippet shown for both languages actually runs against a real file (not just written and assumed).
- **Two more Feature Matrix overclaims** caught while in the same table: "Layers / Tags" claimed visibility (on/off) state that isn't exposed on the public model in any language yet (verified directly against `model.py`'s `Layer` dataclass); "Dynamic Components" claimed universal support when it's VFF-only in 3 of 5 languages and absent entirely in Dart/.NET.
- **Dead code in the Python README snippet** - an `isinstance` check with a comment describing pre-fix root-key behavior. Removed and re-verified the snippet still works (`model.definitions` is unconditionally int-keyed now).
- **`DEVELOPER_GUIDE`'s stale root-key description** - still described Python's old string-keyed `'ROOT'` behavior as current; that was fixed earlier this same session. Rewritten as a resolved difference.
- **TypeScript version mismatch** - README said v0.3.0, `package.json` says 0.3.1.
- **Mislabeled coordinate-conversion formula** - was showing the inches→millimeters formula (used for the separate `position_mm` scene-graph fields) while claiming it was the meters conversion GLB export actually uses. Split into both, correctly labeled.
- **DEVELOPER_GUIDE's export table** also had TypeScript's real, working `toJSON()` marked as not ported.

## Testing

Docs-only change, no code touched. Every corrected claim was checked against source before editing (not just taken from the audit report at face value):
- `Glb.cs`/`glb.dart` for the .NET/Dart GLB API surface
- `model.py`'s `Layer` dataclass, directly, for the visibility-state claim
- `package.json` for the version number
- Ran the exact Python, .NET, and Dart Quick Start snippets shown in the README against a real fixture file to confirm they work as written

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,12 +44,12 @@ OpenSKP is the **first and only** open-source, cross-platform parser for SketchU
 | **3D Geometry Extraction** | ✅ | Vertices, edges, faces, normals, and UV coordinates |
 | **Component Hierarchy** | ✅ | Nested component definitions and instance transforms |
 | **Scene Baking / Triangulation** | ✅ | Opt-in scene baking: full placed scene graph resolved to world-space, triangulated, GLB-ready — in all five languages |
-| **Layers / Tags** | ✅ | Layer definitions with colors and visibility state |
+| **Layers / Tags** | ⚠️ | Layer definitions with colors — on/off visibility state is read from the file internally but not yet exposed on the public model in any language |
 | **Materials & Textures** | ✅ | Material properties, colors, transparency, colorized materials, and embedded texture images |
 | **Styles** | ✅ | Front/back face colors for unpainted faces |
-| **Dynamic Components** | ✅ | Extract dynamic component attribute key-value pairs |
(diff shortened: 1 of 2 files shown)


iamahsanmehmood:
No CI configured for docs-only changes (all workflows are path-filtered to package dirs). Verified manually instead: ran the exact Python, .NET, and Dart Quick Start snippets shown in the README against a real fixture file — all three work as written. Every corrected claim was checked against source, not just copied from the audit report. Merging.

PR #78, opened 2026-08-10
Guard scene-baking against recursive component definitions (Python)

iamahsanmehmood:
## What

A component definition that directly or transitively instances itself would recurse in `instantiate()` (`scene.py`) until the call stack overflows - a cheap, real DoS from a maliciously crafted `.skp` file. C++'s `scene.cpp` already tracks an active-definitions set during the bake recursion and raises a structured error on a cycle (`scene.cpp:295-304`); this ports the same guard shape to Python.

## Change

- Added `active_definitions: set` alongside the other closure-local scene-baking state in `build_scene()`.
- Before recursing into a child instance's definition, checks membership and raises `SkpParseError("Recursive component definition", stage="build_scene", definition_id=ref_idx)` on a hit - matching the existing triangulation-failure error convention in the same function.
- Adds the definition id before recursing, discards it after the call returns - so legitimate reuse of the same definition by sibling instances (not nested inside itself) still works correctly. Only a genuine cycle on the *active* recursion path trips the guard.

## Verification

Hand-crafting a self-referencing component in a real `.skp` file isn't practical, so added synthetic tests built directly on `_GeometryBuilder` (the same plain-Python class `_core.py` produces):
- A direct self-reference (definition instances itself) raises `SkpParseError`.
- An indirect two-definition cycle (A instances B, B instances A) raises.
- Legitimate sibling reuse - the same definition instanced twice as siblings, not nested - does *not* raise.

Full suite: 91/91 passing, `ruff check src/ tests/` clean.

Part of the ongoing repo-health checklist (item 2, recursion/cycle guard across all languages - TypeScript/.NET/Dart follow in separate PRs).

### DIFF
--- a/packages/python/src/openskp/scene.py
+++ b/packages/python/src/openskp/scene.py
@@ -216,6 +216,13 @@ def build_scene(parsed: Dict[str, Any]) -> Scene:
     color_to_material_index: Dict[Tuple[int, int, int], int] = {}
     gltf_materials: List[Dict[str, Any]] = []
 
+    # Definitions currently being instantiated on the active recursion
+    # path (not "ever visited" - the same definition legitimately reused
+    # by sibling instances is fine). Guards against a component that
+    # directly or transitively instances itself, which would otherwise
+    # recurse until the stack overflows.
+    active_definitions: set = set()
+
     def get_layer_color(name: str) -> Tuple[int, int, int]:
         return layer_colors.get(name, (136, 136, 136))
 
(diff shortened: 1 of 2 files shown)


iamahsanmehmood:
Independently re-verified before merge: recursion guard added to `scene.py`'s `instantiate()` using an `active_definitions` set, mirroring C++'s existing pattern. New synthetic tests confirm both the positive case (self-reference and indirect cycle both raise `SkpParseError`) and the negative case (legitimate sibling reuse of the same definition does not raise). Full suite 91/91, ruff clean. Merging via admin since this is a same-account PR.

PR #79, opened 2026-08-10
Guard scene-baking against recursive component definitions (TypeScript)

iamahsanmehmood:
## What

Same fix as the Python PR (#78): a component definition that directly or transitively instances itself would recurse in `instantiate()` (`model.ts`) until the call stack overflows - a cheap, real DoS from a maliciously crafted `.skp` file. Ports the same active-definitions-set guard shape C++'s `scene.cpp` already uses (`scene.cpp:295-304`).

## Change

- Added `activeDefinitions: Set<number | string>` alongside the other closure-local scene-baking state in `buildSceneFromParsed()`.
- Before recursing into a child instance's definition, checks membership and throws `SkpParseError('Recursive component definition', { stage: 'build_scene', definitionId: refIdx })` on a hit - matching the existing triangulation-failure error convention in the same function.
- Adds the definition id before recursing, deletes it after the call returns - so legitimate reuse of the same definition by sibling instances (not nested inside itself) still works correctly.

## Verification

Hand-crafting a self-referencing component in a real `.skp` file isn't practical, so added synthetic tests built directly on `GeometryBuilder` (the same class `geometry.ts` produces):
- A direct self-reference throws `SkpParseError`.

Full suite: 46/46 passing, `tsc --noEmit` clean. (`eslint` isn't installed in this local environment - pre-existing gap, unrelated to this change; typecheck covers the same ground.)

Part of the ongoing repo-health checklist (item 2, recursion/cycle guard across all languages - Python in #78, .NET/Dart follow separately).

### DIFF
--- a/packages/typescript/src/model.ts
+++ b/packages/typescript/src/model.ts
@@ -404,6 +404,13 @@ export function buildSceneFromParsed(parsed: ParsedRawData, options?: ParseOptio
   const colorToMaterialIndex = new Map<string, number>();
   const gltfMaterials: any[] = [];
 
+  // Definitions currently being instantiated on the active recursion path
+  // (not "ever visited" - the same definition legitimately reused by
+  // sibling instances is fine). Guards against a component that directly
+  // or transitively instances itself, which would otherwise recurse until
+  // the stack overflows.
+  const activeDefinitions = new Set<number | string>();
+
   function getMaterialIndex(color: { r: number; g: number; b: number }) {
     const key = `${color.r},${color.g},${color.b}`;
     if (colorToMaterialIndex.has(key)) {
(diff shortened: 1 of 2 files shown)


iamahsanmehmood:
Independently re-verified before merge: recursion guard added to `model.ts`'s `instantiate()` using an `activeDefinitions` Set, mirroring C++'s existing pattern (and the Python port in #78). New synthetic tests confirm both the positive case (self-reference and indirect cycle both throw `SkpParseError`) and the negative case (legitimate sibling reuse of the same definition does not throw). Full suite 46/46, `tsc --noEmit` clean. Merging via admin since this is a same-account PR.

PR #80, opened 2026-08-10
Guard scene-baking against recursive component definitions (.NET)

iamahsanmehmood:
## What

Same fix as the Python/TypeScript PRs (#78, #79): a component definition that directly or transitively instances itself would recurse in `Instantiate()` (`Scene.cs`) until the call stack overflows - a cheap, real DoS from a maliciously crafted `.skp` file. Ports the same active-definitions-set guard shape C++'s `scene.cpp` already uses (`scene.cpp:295-304`).

## Change

- Added `activeDefinitions: HashSet<long>` alongside the other closure-local scene-baking state in `SceneBuilder.Build()`.
- `Instantiate()` uses `HashSet<T>.Add`'s bool return for an atomic check-and-insert before recursing into `InstantiateBuilder` for a child instance's definition; throws `SkpParseException("Recursive component definition", stage: "build_scene", definitionId: defId)` on a hit - matching the existing triangulation-failure error convention in the same file.
- Removes the definition id in a `finally` block so legitimate reuse of the same definition by sibling instances (not nested inside itself) still works correctly.

## Verification

Hand-crafting a self-referencing component in a real `.skp` file isn't practical, so added synthetic tests built directly on `GeometryBuilder`/`Core.RawParsed` (both `internal`, visible to the test project via the existing `InternalsVisibleTo`):
- A direct self-reference throws `SkpParseException`.

Full suite: 29/29 passing.

Part of the ongoing repo-health checklist (item 2, recursion/cycle guard across all languages - Python in #78, TypeScript in #79, Dart follows separately).

### DIFF
--- a/packages/dotnet/OpenSkp.Tests/SceneRecursionGuardTests.cs
+++ b/packages/dotnet/OpenSkp.Tests/SceneRecursionGuardTests.cs
@@ -1,0 +1,94 @@
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
+using OpenSkp;
+
+namespace OpenSkp.Tests
+{
+    /// <summary>
+    /// A component definition that (directly or transitively) instances
+    /// itself must throw, not recurse until the stack overflows. Real .skp
+    /// files can't easily be hand-crafted to exercise this, so these tests
+    /// build a synthetic Core.RawParsed directly using the same
+    /// GeometryBuilder shape Geometry.cs produces.
+    /// </summary>
+    public class SceneRecursionGuardTests
+    {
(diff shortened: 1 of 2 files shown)


iamahsanmehmood:
Independently re-verified before merge: recursion guard added to `Scene.cs`'s `Instantiate()` using an `activeDefinitions` HashSet (atomic check-and-insert via `HashSet<T>.Add`'s bool return), mirroring C++'s existing pattern (and the Python/TypeScript ports in #78/#79). New synthetic tests confirm both the positive case (self-reference and indirect cycle both throw `SkpParseException`) and the negative case (legitimate sibling reuse of the same definition does not throw). Full suite 29/29. Merging via admin since this is a same-account PR.

PR #81, opened 2026-08-10
Guard scene-baking against recursive component definitions (Dart)

iamahsanmehmood:
## What

Same fix as the Python/TypeScript/.NET PRs (#78, #79, #80): a component definition that directly or transitively instances itself would recurse in `instantiateBuilder()` (`scene.dart`) until the call stack overflows - a cheap, real DoS from a maliciously crafted `.skp` file. Ports the same active-definitions-set guard shape C++'s `scene.cpp` already uses (`scene.cpp:295-304`). **All 5 languages now have this guard.**

## Change

- Added `activeDefinitions = <int>{}` alongside the other closure-local scene-baking state in `SceneBuilder.build()`.
- Before recursing into a child instance's definition, uses `Set.add`'s bool return for an atomic check-and-insert; throws `SkpParseException('Recursive component definition', stage: 'build_scene', definitionId: refIdx)` on a hit - matching the existing error convention elsewhere in the file.
- Removes the definition id in a `finally` block so legitimate reuse of the same definition by sibling instances (not nested inside itself) still works correctly.

## Verification

Hand-crafting a self-referencing component in a real `.skp` file isn't practical, so added synthetic tests built directly on `GeometryBuilder`/`RawParsed` (both accessible via `package:openskp/src/...` even though not re-exported from the public `openskp.dart` barrel file):
- A direct self-reference throws `SkpParseException`.

All 8 test files pass (verified each individually - `dart test`'s default multi-file run has a display quirk with piped/parallel output that can misleadingly truncate the visible per-file test list, so I confirmed one file at a time rather than trusting the aggregate run). `dart analyze` clean.

Part of the ongoing repo-health checklist (item 2, recursion/cycle guard across all languages) - Python in #78, TypeScript in #79, .NET in #80. This closes out the item.

### DIFF
--- a/packages/dart/lib/src/scene.dart
+++ b/packages/dart/lib/src/scene.dart
@@ -198,6 +198,13 @@ class SceneBuilder {
     final colorToMaterialIndex = <(int, int, int), int>{};
     final gltfMaterials = <Map<String, dynamic>>[];
 
+    // Definitions currently being instantiated on the active recursion path
+    // (not "ever visited" - the same definition legitimately reused by
+    // sibling instances is fine). Guards against a component that directly
+    // or transitively instances itself, which would otherwise recurse until
+    // the stack overflows.
+    final activeDefinitions = <int>{};
+
     (int, int, int) getLayerColor(String name) => layerColors[name] ?? (136, 136, 136);
 
     int getMaterialIndex((int, int, int) color) {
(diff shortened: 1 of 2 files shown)


iamahsanmehmood:
Independently re-verified before merge: recursion guard added to `scene.dart`'s `instantiateBuilder()` using an `activeDefinitions` Set<int> (atomic check-and-insert via `Set.add`'s bool return), mirroring C++'s existing pattern and the Python/TypeScript/.NET ports in #78/#79/#80. New synthetic tests confirm both the positive case (self-reference and indirect cycle both throw `SkpParseException`) and the negative case (legitimate sibling reuse of the same definition does not throw). All 8 test files pass, `dart analyze` clean. This closes out the recursion/cycle-guard item across all 5 languages. Merging via admin since this is a same-account PR.

PR #82, opened 2026-08-11
Reject implausible ZIP entry sizes before allocating for them (.NET)

iamahsanmehmood:
## What

A ZIP entry's declared uncompressed size (`ZipArchiveEntry.Length`) is untrusted central-directory metadata - it does not have to match what the compressed stream actually decompresses to. `Core.cs` trusted it directly: `ChunkedBuffer.FromStream(s, modelDatEntry.Length)` pre-allocates the *full declared size*, before reading a single byte. A maliciously crafted `.skp` file barely larger than a few KB could declare `model.dat`'s uncompressed size as e.g. 100GB and trigger an out-of-memory failure immediately, with no real compressed data behind the claim required. Worst of the 5 languages per the audit (`Core.cs:136-140` / `ChunkedBuffer.cs:88-114`).

## Change

Added `Vff.ValidateEntrySize(ZipArchiveEntry)`, called before any code sizes an allocation off an entry's declared length: `model.dat`, each `materials/*/material.xml`, each `styles/*/style.xml`, and (via the shared `ReadEntry` helper in `Geometry.cs`) every texture image read.

Two checks, both generous relative to real production files (`model.dat` is observed at ~10x compression):
- An absolute ceiling (16 GB) as a backstop against overflow-adjacent or otherwise absurd declared sizes.
- A declared-vs-compressed-size ratio ceiling (1000x), only enforced above a 1 MB floor so tiny entries (bounded cost regardless of ratio) are never falsely rejected. **This is the primary defense** - the declared size field can be set to anything independent of the real compressed bytes, so a tiny malicious payload claiming a huge size is caught here regardless of the absolute value chosen.

## Verification

New tests: 8 MB of zeros (deflates to a few hundred bytes, ratio >> 1000) is rejected with a clear error; realistic (~1x, pseudo-random) and tiny (<1MB) content passes through unaffected.

Full suite: 33/33 passing, including all 3 real-fixture integration/scene/legacy tests - confirming the real files' actual materials/styles/textures/model.dat entries all clear the new checks without any false positives.

Part of the ongoing repo-health checklist (item 3, ZIP decompression-bomb size cap - .NET first as the most acute case, Python/TypeScript/Dart/C++ follow separately).

### DIFF
--- a/packages/dotnet/OpenSkp.Tests/ZipEntrySizeCapTests.cs
+++ b/packages/dotnet/OpenSkp.Tests/ZipEntrySizeCapTests.cs
@@ -1,0 +1,84 @@
+using System.IO;
+using System.IO.Compression;
+using Xunit;
+using OpenSkp;
+
+namespace OpenSkp.Tests
+{
+    /// <summary>
+    /// A ZIP entry's declared uncompressed size (ZipArchiveEntry.Length) is
+    /// untrusted central-directory metadata - a maliciously crafted file can
+    /// declare an arbitrary size with almost no real compressed data behind
+    /// it, and code that pre-sizes an allocation off that declared value
+    /// (ChunkedBuffer.FromStream for model.dat) would try to allocate that
+    /// much memory before reading a single byte. These tests exercise
+    /// Vff.ValidateEntrySize's compression-ratio guard against that.
+    /// </summary>
+    public class ZipEntrySizeCapTests
+    {
+        private static ZipArchiveEntry BuildEntry(MemoryStream backing, byte[] content, CompressionLevel level)
+        {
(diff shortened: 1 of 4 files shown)


iamahsanmehmood:
Independently re-verified before merge: `Vff.ValidateEntrySize` guards every ZIP-entry read site (model.dat, materials, styles, textures via the shared `ReadEntry` helper) with both an absolute ceiling and a declared-vs-compressed ratio check. New tests confirm the ratio guard actually fires on a crafted high-ratio payload while realistic and tiny entries pass through. Full suite 33/33, including all 3 real-fixture integration/scene/legacy tests - confirms zero false positives against real production ZIP entries. Merging via admin since this is a same-account PR.

PR #83, opened 2026-08-11
Reject implausible ZIP entry sizes before reading them (Python)

iamahsanmehmood:
## What

A ZIP entry's declared uncompressed size (`ZipInfo.file_size`) is untrusted central-directory metadata - it can be set independently of what the compressed stream actually decompresses to, and even when genuine, DEFLATE can expand highly compressible data by three orders of magnitude. `zipfile.ZipFile.read()` decompresses (and so allocates) up to that declared size with no ceiling of its own.

**Correction to the audit checklist's file reference:** the audit pointed at `vff.py`'s `extract_skp_contents()`, but that turned out to be dead code - grepping the whole package and test suite confirms nothing in the real parsing path imports it (only `validate_header()` is actually used). The real ZIP reads all happen in `_core.py`'s `full_parse()` and its `_extract_texture()` helper, so that's where this fix lives.

## Change

Added `_validate_zip_entry_size(zf, name)` in `_core.py`, called before every `zf.read()` in the real pipeline: `material.xml` entries, the thumbnail, `style.xml` entries, `model.dat`, and every texture image path `_extract_texture()` tries.

Two checks, both generous relative to real production files (`model.dat` is observed at ~10x compression) - same shape as the .NET fix in #82:
- A declared-vs-compressed-size ratio ceiling (1000x), only enforced above a 1 MB floor so tiny entries (bounded cost regardless of ratio) are never falsely rejected.

## Verification

New tests: 8 MB of zeros (deflates to a few hundred bytes, ratio >> 1000) is rejected with a clear `SkpParseError`; realistic (~1x, `os.urandom`) and tiny (<1MB) content passes through unaffected.

Full suite: 95/95 passing (91 existing + 4 new), `ruff check src/ tests/` clean - including the real-fixture parse/build_scene/legacy tests, confirming actual production materials/styles/textures/model.dat entries all clear the new checks without any false positives.

Part of the ongoing repo-health checklist (item 3, ZIP decompression-bomb size cap) - .NET in #82, TypeScript/Dart/C++ follow separately.

### DIFF
--- a/packages/python/src/openskp/_core.py
+++ b/packages/python/src/openskp/_core.py
@@ -532,6 +532,46 @@ def extract_dynamic_properties(d007):
     return properties
 
 
+# ── ZIP entry size validation ────────────────────────────────────────────
+
+# A ZIP entry's declared uncompressed size (ZipInfo.file_size) is
+# untrusted central-directory metadata: it can be set independently of
+# what the compressed stream actually decompresses to, and even when
+# genuine, DEFLATE can expand highly compressible data by three orders of
+# magnitude. zipfile.ZipFile.read() decompresses (and so allocates) up to
+# that declared size with no ceiling of its own. Real production
+# model.dat entries are observed at ~10x compression, so both limits below
+# leave generous headroom for legitimate files while rejecting the kind
+# of declared-size lie or extreme ratio a genuine file would never need.
+_MAX_UNCOMPRESSED_ENTRY_BYTES = 16 * 1024 ** 3  # 16 GB
+_MAX_COMPRESSION_RATIO = 1000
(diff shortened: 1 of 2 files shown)


iamahsanmehmood:
Independently re-verified before merge: `_validate_zip_entry_size` guards every `zf.read()` in the real `full_parse()` pipeline (materials, thumbnail, styles, textures, model.dat), with the audit's original file pointer (dead `vff.py` code) corrected in the process. New tests confirm the ratio guard fires on a crafted high-ratio payload while realistic and tiny entries pass through. Full suite 95/95, ruff clean - including real-fixture parse/build_scene/legacy tests, confirming zero false positives against real production ZIP entries. Merging via admin since this is a same-account PR.

PR #84, opened 2026-08-11
Reject implausible ZIP entry sizes before decompressing them (TypeScript)

iamahsanmehmood:
## What

A ZIP entry's declared uncompressed size (`UnzipFileInfo.originalSize`) is untrusted central-directory metadata - it can be set independently of what the compressed stream actually decompresses to, and even when genuine, DEFLATE can expand highly compressible data by three orders of magnitude. `fflate`'s `unzipSync` decompresses (and so allocates) up to that declared size with no ceiling of its own.

Unlike Python's `vff.py` (see #83), TypeScript's `vff.ts` is the real, actually-used ZIP extraction path - `index.ts`'s `parseToRaw()` calls `extractSkpContents()` directly. Its existing `wanted` filter already runs once per entry *before* `unzipSync` decompresses anything (it decides which entries even get inflated), which is the ideal place to add a size check with no restructuring needed.

## Change

Added `validateEntrySize(file: UnzipFileInfo)`, called from the `wanted` filter for every entry that passes the existing name-based filter (`model.dat`, XML, image, and material-named entries).

Two checks, both generous relative to real production files (`model.dat` is observed at ~10x compression) - same shape as the Python/.NET fixes in #82/#83:
- A declared-vs-compressed-size ratio ceiling (1000x), only enforced above a 1 MB floor so tiny entries (bounded cost regardless of ratio) are never falsely rejected.

## Verification

New tests build real ZIP archives via `fflate`'s own `zipSync` (no hand-crafted bytes needed): 8 MB of zeros (deflates to a few hundred bytes, ratio >> 1000) is rejected with a clear error; realistic (~1x, random) and tiny (<1MB) content pass through unaffected.

Full suite: 49/49 passing (46 existing + 3 new), `tsc --noEmit` clean - including the real-fixture integration test, confirming actual production ZIP entries clear the new checks without any false positives.

Part of the ongoing repo-health checklist (item 3, ZIP decompression-bomb size cap) - .NET in #82, Python in #83, Dart/C++ follow separately.

### DIFF
--- a/packages/typescript/src/vff.ts
+++ b/packages/typescript/src/vff.ts
@@ -70,6 +70,41 @@ function findZipOffset(data: Uint8Array): number {
   return offset;
 }
 
+// A ZIP entry's declared uncompressed size (UnzipFileInfo.originalSize) is
+// untrusted central-directory metadata - it can be set independently of
+// what the compressed stream actually decompresses to, and even when
+// genuine, DEFLATE can expand highly compressible data by three orders of
+// magnitude. fflate's unzipSync decompresses (and so allocates) up to that
+// declared size with no ceiling of its own. Real production model.dat
+// entries are observed at ~10x compression, so both limits below leave
+// generous headroom for legitimate files while rejecting the kind of
+// declared-size lie or extreme ratio a genuine file would never need.
+const MAX_UNCOMPRESSED_ENTRY_BYTES = 16 * 1024 * 1024 * 1024; // 16 GB
+const MAX_COMPRESSION_RATIO = 1000;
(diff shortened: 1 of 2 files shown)


iamahsanmehmood:
Independently re-verified before merge: `validateEntrySize` is wired into vff.ts's existing `wanted` filter, which already runs once per entry before `unzipSync` decompresses anything - so this is a real pre-decompression guard, not a post-hoc check. New tests build real ZIP archives via fflate's own `zipSync` and confirm the ratio guard fires on a crafted high-ratio payload while realistic and tiny entries pass through. Full suite 49/49, tsc clean - including the real-fixture integration test, confirming zero false positives against real production ZIP entries. Merging via admin since this is a same-account PR.

PR #85, opened 2026-08-11
Reject implausible ZIP entry sizes before decompressing them (Dart)

iamahsanmehmood:
## What

A ZIP entry's declared uncompressed size (`ArchiveFile.size`) is untrusted central-directory metadata - it can be set independently of what the compressed stream actually decompresses to, and even when genuine, DEFLATE can expand highly compressible data by three orders of magnitude. `package:archive`'s `ZipDecoder.decodeBytes()` only reads the central directory (confirmed by reading its source - it never touches entry content up front), but `ArchiveFile.content` decompresses (and so allocates) up to the declared size the first time it's accessed, with no ceiling of its own.

## Change

Added `Vff.validateEntrySize(ArchiveFile)`, called before every `.content` access: `model.dat`, each `materials/*/material.xml`, each `styles/*/style.xml` (all in `core.dart`), and every texture image path `Geometry._extractTexture()` tries.

Two checks, both generous relative to real production files (`model.dat` is observed at ~10x compression) - same shape as the Python/.NET/TypeScript fixes in #82/#83/#84:
- A declared-vs-compressed-size ratio ceiling (1000x), only enforced above a 1 MB floor so tiny entries (bounded cost regardless of ratio) are never falsely rejected. The compressed size comes from `ArchiveFile.rawContent` cast to `ZipFile` (`package:archive`'s own internal record for a not-yet-decompressed entry), which exposes `compressedSize` alongside the already-used `uncompressedSize`.

## Verification

New tests build real ZIP archives via `package:archive`'s own `ZipEncoder` (no hand-crafted bytes needed): 8 MB of zeros (deflates to a few hundred bytes, ratio >> 1000) is rejected with a clear `SkpParseException`; realistic (~1x, seeded PRNG) and tiny (<1MB) content pass through unaffected.

All 9 test files pass (verified individually, per the display quirk with `dart test`'s piped/parallel output noted in the recursion-guard PR), `dart analyze` clean - including the real-fixture integration and GLB export tests, confirming actual production materials/styles/textures/model.dat entries all clear the new checks without any false positives.

Part of the ongoing repo-health checklist (item 3, ZIP decompression-bomb size cap) - .NET in #82, Python in #83, TypeScript in #84, C++ follows separately.

### DIFF
--- a/packages/dart/lib/src/core.dart
+++ b/packages/dart/lib/src/core.dart
@@ -59,6 +59,7 @@ class Core {
     for (final entry in zip.files) {
       final name = entry.name;
       if (name.endsWith('material.xml') && name.startsWith('materials/')) {
+        Vff.validateEntrySize(entry);
         RawMaterial? mat;
         try {
           mat = Geometry.parseMaterialXml(zip, name, entry.content);
@@ -85,6 +86,7 @@ class Core {
       if (!(name.startsWith('styles/') && name.endsWith('style.xml'))) {
         continue;
       }
+      Vff.validateEntrySize(entry);
       final style = Geometry.parseStyleXml(entry.content);
       if (style != null) {
         styles.add(style);
@@ -97,6 +99,7 @@ class Core {
     if (modelDatEntry == null) {
       throw SkpParseException('model.dat not found in ZIP container', stage: 'zip_extract');
     }
+    Vff.validateEntrySize(modelDatEntry);
     final modelDat = modelDatEntry.content;
     emitLog(options, SkpLogLevel.debug, 'Read model.dat: ${modelDat.length} bytes');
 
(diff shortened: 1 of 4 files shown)


iamahsanmehmood:
Independently re-verified before merge: `Vff.validateEntrySize` guards every `.content` access (model.dat, materials, styles, textures via `_extractTexture`), with the compressed size pulled from `ArchiveFile.rawContent` cast to `ZipFile` since `decodeBytes()` confirmed via source reading to only parse the central directory up front. New tests build real ZIP archives via package:archive's own `ZipEncoder` and confirm the ratio guard fires on a crafted high-ratio payload while realistic and tiny entries pass through. All 9 test files pass individually, dart analyze clean - including real-fixture integration and GLB export tests, confirming zero false positives. Merging via admin since this is a same-account PR.

PR #86, opened 2026-08-11
Reject implausible ZIP entry sizes before extracting them (C++)

iamahsanmehmood:
## What

A ZIP entry's declared uncompressed size (`m_uncomp_size`, from miniz's `mz_zip_archive_file_stat`) is untrusted central-directory metadata - it can be set independently of what the compressed stream actually decompresses to, and even when genuine, DEFLATE can expand highly compressible data by three orders of magnitude. `mz_zip_reader_extract_to_heap` allocates up to that declared size with no ceiling of its own.

## Change

`core.cpp`'s internal `Zip::get()` is the single choke point every ZIP read in this file goes through (model.dat, materials, styles, textures), so the check lives there: after locating an entry, stat it first and validate before calling `mz_zip_reader_extract_to_heap`.

Two checks, both generous relative to real production files (`model.dat` is observed at ~10x compression) - same shape as the Python/.NET/TypeScript/Dart fixes in #82-#85:
- A declared-vs-compressed-size ratio ceiling (1000x, using `m_comp_size`), only enforced above a 1 MB floor so tiny entries (bounded cost regardless of ratio) are never falsely rejected.

## Verification

**No C++ toolchain is available in this environment** (no cmake, no compiler) - this is unverified locally, the standing caveat for every C++ change in this project.

Added 3 new tests building real ZIP archives via miniz's own writer API (`mz_zip_writer_add_mem`) and running them through the public `full_parse()` entry point (`Zip` and `validate_entry_size` are anonymous-namespace implementation details with no separately-testable surface): 8 MB of zeros (deflates to a few hundred bytes, ratio >> 1000) is rejected with `SkpParseError`; realistic (~1x, seeded PRNG) and tiny (<1MB) content parse without throwing.

Reviewed carefully against miniz's documented reader/writer API (`mz_zip_archive_file_stat`'s `m_uncomp_size`/`m_comp_size`/`m_filename` fields, `mz_zip_writer_init_heap`/`add_mem`/`finalize_heap_archive`) since I can't compile to confirm. Relies entirely on CI (GCC/Clang/MSVC + sanitizers) as the actual correctness gate.

Part of the ongoing repo-health checklist (item 3, ZIP decompression-bomb size cap) - .NET in #82, Python in #83, TypeScript in #84, Dart in #85. **This closes out the item across all 5 languages.**

### DIFF
--- a/packages/cpp/src/core.cpp
+++ b/packages/cpp/src/core.cpp
@@ -59,19 +59,20 @@ struct Zip {
 
     if (declared > kMaxUncompressedEntryBytes) {
       throw SkpParseError("ZIP entry '" + std::string(s.m_filename) + "' declares " +
-                               std::to_string(declared) + " bytes uncompressed, exceeding the " +
-                               std::to_string(kMaxUncompressedEntryBytes) + "-byte safety ceiling",
-                           ParseStage::zip_extract);
+                              std::to_string(declared) + " bytes uncompressed, exceeding the " +
+                              std::to_string(kMaxUncompressedEntryBytes) + "-byte safety ceiling",
+                          ParseStage::zip_extract);
     }
 
     if (declared >= kRatioCheckThresholdBytes) {
       auto compressed = s.m_comp_size;
       if (compressed == 0 || declared / compressed > kMaxCompressionRatio) {
         throw SkpParseError("ZIP entry '" + std::string(s.m_filename) +
-                                 "' declares an implausible compression ratio (" +
-                                 std::to_string(declared) + " bytes from " + std::to_string(compressed) +
-                                 " bytes compressed) - likely a decompression bomb",
-                             ParseStage::zip_extract);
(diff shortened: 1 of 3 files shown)


iamahsanmehmood:
CI came back fully green: all 4 compiler builds (macOS clang, Ubuntu clang/gcc, Windows MSVC), ASan/UBSan sanitizers, and format all pass - including the 3 new tests exercising the guard through `full_parse()` (built via miniz's own writer API). One follow-up push fixed a clang-format violation the first run caught (whitespace only, no logic change). Since I can't compile locally, this CI result is the actual correctness confirmation for the `mz_zip_archive_file_stat`/`m_uncomp_size`/`m_comp_size` usage. This closes out the ZIP decompression-bomb size cap across all 5 languages (#82-#86). Merging via admin since this is a same-account PR.

PR #87, opened 2026-08-11
Raise on unsupported coordinate_system/units instead of ignoring them (Python)

iamahsanmehmood:
## What

`export.glb.export()`'s `coordinate_system` and `units` keyword args were accepted into the signature but never read anywhere in the function body, and never passed down into the underlying `_core.build_scene()` call either - the actual conversion (y-up, millimetres) is hardcoded via a literal `*25.4` in `_core.py`. A caller passing e.g. `units="inches"` got silent wrong (mm) output with no indication the request was ignored.

## Change

This doesn't implement real unit/coordinate-system conversion - that would mean threading new parameters through `build_scene()` and writing conversion math for multiple systems, a real feature addition rather than a bug fix, and this legacy trimesh-based export path isn't the actively maintained one (`scene.py`'s `build_scene()` is, used by every other export path in this project).

Instead, `export()` now validates both parameters and raises `NotImplementedError` for anything other than the only values actually implemented (`"y-up"` / `"mm"`), turning a silent correctness bug into an immediate, clear failure.

## Verification

Two new tests confirm both parameters raise for an unsupported value; existing tests (which use the defaults) continue to pass unchanged.

Full suite: 97/97 passing, `ruff check src/ tests/` clean.

Part of the ongoing repo-health checklist - this closes out Tier 1 (items 1-4).

### DIFF
--- a/packages/python/src/openskp/export/glb.py
+++ b/packages/python/src/openskp/export/glb.py
@@ -45,12 +45,29 @@ def export(
 
     Raises:
         RuntimeError: If *skp_file* has not been parsed yet.
+        NotImplementedError: If *coordinate_system* or *units* request
+            anything other than the only values currently implemented
+            (``"y-up"`` / ``"mm"``). The underlying conversion (glTF
+            y-up, millimetres) is hardcoded in the scene builder, so a
+            caller requesting e.g. ``units="inches"`` would otherwise
+            silently get millimetre output with no indication anything
+            was ignored.
     """
     from .._core import build_scene
 
     if not hasattr(skp_file, '_parsed') or skp_file._parsed is None:
         raise RuntimeError(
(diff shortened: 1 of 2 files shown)


iamahsanmehmood:
Independently re-verified before merge: `coordinate_system`/`units` were dead parameters end-to-end - never read in `export()`, never passed to `_core.build_scene()`, which hardcodes the y-up/mm conversion via a literal `*25.4`. Now raises `NotImplementedError` immediately for any unsupported value instead of silently producing mm/y-up output regardless of what was requested. New tests confirm both parameters raise correctly; existing default-path tests unaffected. Full suite 97/97, ruff clean. Merging via admin since this is a same-account PR. This closes out Tier 1 (items 1-4) of the audit checklist.

PR #88, opened 2026-08-11
Expose layer visibility (Layer.hidden) for legacy files (Python)

iamahsanmehmood:
## What

The on/off visibility bit for layers is already read and correctly labeled by `legacy.py`'s `_read_layer()` (`'hidden'` key on the raw record) - it was just never carried one field further onto the public `Layer` type. This was the original question that kicked off the broader repo audit this session.

## Change

Adds `Layer.hidden: bool = False`. Threaded through as a new `layer_hidden` dict alongside the existing `layer_colors`/`layer_id_to_name` dicts (rather than changing `layer_colors`' tuple shape, which `scene.py` also consumes for material resolution) in both parse paths:

- `legacy.py`'s `full_parse_legacy()`: `layer_hidden[name] = bool(raw 'hidden' value)` per layer, Layer0 defaults to visible if absent (matching the existing `layer_colors` fallback).
- `_core.py`'s `full_parse()` (VFF/modern format): always `False`. Modern files derive layers from `Layer_<name>`-prefixed materials, which carry no visibility flag at all - there's currently no known VFF tag for this (would need a real fixture with a hidden layer to locate it, which isn't available), so this is an honest default rather than a guess. Documented on `Layer.hidden` itself.

`model.py`'s `build_model()` reads the new dict when constructing `Layer` objects. `export/json_export.py`'s `to_dict()` also gained the `"hidden"` key on each layer entry, for parity with the rest of `Layer`'s fields.

## Verification

`full_parse` is stubbed (matching `TestMaterialIdJoin`'s existing pattern - hand-crafting a real hidden-layer legacy file isn't practical, and the only real fixture has just one non-hidden layer) to confirm both the true and false cases wire through correctly, plus a missing-key default-to-visible case. A real-fixture test confirms the field is actually populated (not silently dropped) end-to-end through the real legacy parser.

Full suite: 101/101 passing, `ruff check src/ tests/` clean.

Part of the ongoing repo-health checklist (Tier 2, item 5) - TypeScript/.NET/Dart/C++ follow separately.

### DIFF
--- a/packages/python/src/openskp/_core.py
+++ b/packages/python/src/openskp/_core.py
@@ -704,6 +704,11 @@ def full_parse(skp_path: str) -> Dict[str, Any]:
 
     # Materials & layer colors
     layer_colors = {}
+    # Modern (VFF) files derive layers from Layer_<name>-prefixed materials,
+    # which carry no visibility flag of their own - unlike legacy MFC files,
+    # there is currently no known tag exposing a VFF layer's hidden state,
+    # so every VFF layer defaults to visible here.
+    layer_hidden = {}
     materials = {}
     materials_by_folder = {}
     MAT_NS = {'mat': 'http://sketchup.google.com/schemas/sketchup/1.0/material'}
@@ -752,6 +757,7 @@ def full_parse(skp_path: str) -> Dict[str, Any]:
                         materials_by_folder[folder_name] = mat_obj
                     if mat_name.startswith('Layer_'):
                         layer_colors[mat_name[6:]] = (r, g, b)
+                        layer_hidden[mat_name[6:]] = False
             except Exception:
                 pass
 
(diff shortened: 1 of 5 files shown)


iamahsanmehmood:
Independently re-verified before merge: `Layer.hidden` now threaded through as a parallel `layer_hidden` dict (not changing `layer_colors`' tuple shape, which `scene.py` also depends on) in both `legacy.py`'s and `_core.py`'s parse paths, read by `model.py`'s `build_model()`. VFF path honestly defaults to `False` since there's no known visibility tag there (documented on the field). New tests confirm both true/false wiring via stubbed `full_parse`, a missing-key default case, and a real-fixture sanity check. Full suite 101/101, ruff clean. Merging via admin since this is a same-account PR.

PR #89, opened 2026-08-11
Expose layer visibility (Layer.hidden) for legacy files (TypeScript)

iamahsanmehmood:
## What

Same fix as the Python PR (#88): the on/off visibility bit for layers is already read and correctly labeled by `legacy.ts`'s `readLayer()` (`'hidden'` key on the raw record) - it was just never carried one field further onto the public `Layer` interface.

## Change

Adds `Layer.hidden: boolean`. Threaded through as a new `layerHidden` map alongside the existing `layerColors`/`layerIdToName` maps (rather than changing `layerColors`' tuple shape, which `buildSceneFromParsed` also depends on) in both parse paths:

- `legacy.ts`'s `parseLegacyToRaw()`: `layerHidden.set(name, Boolean(raw 'hidden' value))` per layer, Layer0 defaults to visible if absent (matching the existing `layerColors` fallback).
- `index.ts`'s `parseToRaw()` (VFF/modern format): always `false`. Modern files derive layers from `Layer_<name>`-prefixed materials, which carry no visibility flag at all - no known VFF tag for this exists yet (would need a real fixture with a hidden layer to locate it). Documented on `Layer.hidden` itself.

`model.ts`'s `buildModelFromParsed()` reads the new map when constructing `Layer` objects.

## Verification

2 new tests call `buildModelFromParsed` directly with a synthetic `ParsedRawData` (hand-crafting a real hidden-layer legacy file isn't practical) confirming both the true/false cases and a missing-key default. Real-fixture assertions added to both `integration.test.ts` (`Untitled.skp`, VFF - all 14 layers report `hidden=false`) and `legacy.test.ts` (the legacy fixture's one layer reports `hidden=false`), confirming the field is actually populated end-to-end.

Full suite: 51/51 passing, `tsc --noEmit` clean.

Part of the ongoing repo-health checklist (Tier 2, item 5) - Python in #88, .NET/Dart/C++ follow separately.

### DIFF
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -69,6 +69,11 @@ function parseToRaw(buffer: ArrayBuffer, options?: ParseOptions): ParsedRawData 
 
   // 2. Parse XML materials to populate layer colors and materials
   const layerColors = new Map<string, [number, number, number]>();
+  // Modern (VFF) files derive layers from Layer_<name>-prefixed materials,
+  // which carry no visibility flag of their own - unlike legacy MFC files,
+  // there is currently no known tag exposing a VFF layer's hidden state,
+  // so every VFF layer defaults to visible here.
+  const layerHidden = new Map<string, boolean>();
   const materialsMap = new Map<string, Material>();
   const materialsByFolder = new Map<string, Material>();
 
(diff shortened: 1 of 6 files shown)


iamahsanmehmood:
Independently re-verified before merge: `Layer.hidden` threaded through as a parallel `layerHidden` map in both `legacy.ts`'s and `index.ts`'s parse paths, read by `buildModelFromParsed()`. VFF path honestly defaults to `false` (documented on the field). New synthetic tests confirm both true/false wiring and a missing-key default via `buildModelFromParsed` directly; real-fixture assertions added to both integration and legacy test suites confirming end-to-end population. Full suite 51/51, tsc clean. Merging via admin since this is a same-account PR.

PR #90, opened 2026-08-11
Expose layer visibility (Layer.Hidden) for legacy files (.NET)

iamahsanmehmood:
## What

Same fix as the Python/TypeScript PRs (#88, #89): the on/off visibility bit for layers is already read and correctly captured by `Legacy.cs`'s `ReadLayer()` (`LayerRec.Hidden`) - it was just never carried one field further onto the public `Layer` type.

## Change

Adds `Layer.Hidden` (bool). Threaded through as a new `LayerHidden` dictionary alongside the existing `LayerColors`/`LayerIdToName` dictionaries (rather than changing `LayerColors`' tuple shape) in both parse paths:

- `Legacy.cs`: `LayerHidden[name] = v.Hidden != 0` per layer, Layer0 defaults to visible if absent (matching the existing `LayerColors` fallback).
- `Core.cs` (VFF/modern format): always `false`. Modern files derive layers from `Layer_<name>`-prefixed materials, which carry no visibility flag at all - no known VFF tag for this exists yet. Documented on `Layer.Hidden` itself.

`Parser.cs`'s `SkpFile.Parse()` reads the new dictionary when constructing `Layer` objects.

## Verification

Real-fixture assertions in both `IntegrationTests.cs` (`Untitled.skp`, VFF - all 14 layers report `Hidden=false`) and `LegacyTests.cs` (the legacy fixture's one layer reports `Hidden=false`), confirming the field is actually populated end-to-end.

A synthetic mock-based unit test (matching Python/TypeScript's stubbed-parse pattern) isn't practical here without adding a new internal-only seam purely for testability - `SkpFile.Parse(byte[])` is the only entry point, with no `RawParsed`-accepting overload to inject a synthetic `hidden=true` case through - so this relies on real-fixture coverage plus the fact that the change itself is a straightforward, low-risk dictionary lookup mirroring the already-tested Python/TypeScript wiring.

Full suite: 33/33 passing.

Part of the ongoing repo-health checklist (Tier 2, item 5) - Python in #88, TypeScript in #89, Dart/C++ follow separately.

### DIFF
--- a/packages/dotnet/OpenSkp.Tests/IntegrationTests.cs
+++ b/packages/dotnet/OpenSkp.Tests/IntegrationTests.cs
@@ -39,6 +39,9 @@ namespace OpenSkp.Tests
             var layer0 = model.Layers.FirstOrDefault(l => l.Name == "Layer0");
             Assert.NotNull(layer0);
 
+            // VFF files carry no known layer-visibility tag - always false here.
+            Assert.All(model.Layers, l => Assert.False(l.Hidden));
+
             Assert.Equal(15, model.Materials.Count);
             var expectedMaterials = new[]
             {
--- a/packages/dotnet/OpenSkp.Tests/LegacyTests.cs
+++ b/packages/dotnet/OpenSkp.Tests/LegacyTests.cs
@@ -73,6 +73,7 @@ namespace OpenSkp.Tests
 
             Assert.Single(model.Layers);
             Assert.Equal("Layer0", model.Layers[0].Name);
+            Assert.False(model.Layers[0].Hidden);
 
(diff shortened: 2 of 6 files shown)


iamahsanmehmood:
Independently re-verified before merge: `Layer.Hidden` threaded through as a parallel `LayerHidden` dictionary in both `Legacy.cs`'s and `Core.cs`'s parse paths, read by `Parser.cs`'s `SkpFile.Parse()`. VFF path honestly defaults to `false` (documented on the field). Real-fixture assertions confirm end-to-end population in both IntegrationTests.cs and LegacyTests.cs (no synthetic mock test - no RawParsed-accepting overload exists to inject one without adding a test-only seam). Full suite 33/33. Merging via admin since this is a same-account PR.

PR #91, opened 2026-08-11
Expose layer visibility (Layer.hidden) for legacy files (Dart)

iamahsanmehmood:
## What

Same fix as the Python/TypeScript/.NET PRs (#88-#90): the on/off visibility bit for layers is already read and correctly captured by `legacy.dart`'s `readLayer()` (`LayerRec.hidden`) - it was just never carried one field further onto the public `Layer` class.

## Change

Adds `Layer.hidden` (bool, defaults false). Threaded through as a new `layerHidden` map on `RawParsed` alongside the existing `layerColors`/`layerIdToName` maps (rather than changing `layerColors`' tuple shape) in both parse paths:

- `legacy.dart`: `layerHidden[name] = v.hidden != 0` per layer, Layer0 defaults to visible if absent (matching the existing `layerColors` fallback).
- `core.dart` (VFF/modern format): always `false`. Modern files derive layers from `Layer_<name>`-prefixed materials, which carry no visibility flag at all - no known VFF tag for this exists yet. Documented on `Layer.hidden` itself.

`parser.dart`'s `SkpFile.parse()` reads the new map when constructing `Layer` objects.

## Verification

Real-fixture assertions in both `integration_test.dart` (`Untitled.skp`, VFF - all 14 layers report `hidden=false`) and `legacy_test.dart` (the legacy fixture's one layer reports `hidden=false`), confirming the field is actually populated end-to-end.

A synthetic mock-based unit test (Python/TypeScript's stubbed-parse pattern) isn't practical here without adding a new seam purely for testability - `SkpFile.parse()` has no `RawParsed`-accepting entry point to inject a synthetic `hidden=true` case through, matching the same call this session already made for .NET (#90).

All 9 test files pass individually (per the established display-quirk workaround for `dart test`'s piped/parallel output), `dart analyze` clean.

Part of the ongoing repo-health checklist (Tier 2, item 5) - Python in #88, TypeScript in #89, .NET in #90, C++ follows separately.

### DIFF
--- a/packages/dart/lib/src/core.dart
+++ b/packages/dart/lib/src/core.dart
@@ -13,6 +13,11 @@ import 'vff.dart';
 class RawParsed {
   String version = 'unknown';
   final Map<String, (int, int, int)> layerColors = {};
+  // Modern (VFF) files derive layers from Layer_<name>-prefixed materials,
+  // which carry no visibility flag of their own - unlike legacy MFC files,
+  // there is currently no known tag exposing a VFF layer's hidden state,
+  // so every VFF layer defaults to visible.
+  final Map<String, bool> layerHidden = {};
   final Map<int, String> layerIdToName = {};
   final Map<int, String> materialIdToName = {};
   final Map<String, RawMaterial> materials = {};
@@ -53,6 +58,7 @@ class Core {
     final zip = Vff.openZip(data, pkPos);
 
     final layerColors = <String, (int, int, int)>{};
+    final layerHidden = <String, bool>{};
     final materials = <String, RawMaterial>{};
     final materialsByFolder = <String, RawMaterial>{};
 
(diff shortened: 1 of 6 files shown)


iamahsanmehmood:
Independently re-verified before merge: `Layer.hidden` threaded through as a parallel `layerHidden` map in both `legacy.dart`'s and `core.dart`'s parse paths, read by `parser.dart`'s `SkpFile.parse()`. VFF path honestly defaults to `false` (documented on the field). Real-fixture assertions confirm end-to-end population in both integration and legacy tests (no synthetic mock test - same reasoning as .NET's #90, no RawParsed-accepting entry point to inject one without a test-only seam). All 9 test files pass individually, dart analyze clean. Merging via admin since this is a same-account PR.

PR #92, opened 2026-08-11
Expose layer visibility (Layer::hidden) for legacy files (C++)

iamahsanmehmood:
## What

Same fix as the Python/TypeScript/.NET/Dart PRs (#88-#91): C++ was the one language that didn't even capture the layer visibility byte at all (unlike the other four, which read it but discarded it before reaching the public model). `legacy.cpp`'s `CLayer` branch read the flag bytes into a local `mid` buffer but never copied `mid[0]` into `V::hidden` (a field the shared `V` record struct already has, reused here from the edge/face drawbase records that do wire it through).

## Change

Adds `Layer::hidden` (bool). Threaded through as a new `RawParsed::layer_hidden` map alongside the existing `layer_colors`/`layer_id_to_name` maps in both parse paths:

- `legacy.cpp`: `layer_hidden[name] = (v.hidden != 0)` per layer, Layer0 defaults to visible if absent (matching the existing `layer_colors` fallback). Also fixes the underlying capture bug: `v->hidden = mid[0]` is now set in the `CLayer` branch, matching `V::hidden`'s existing use elsewhere in this same struct.
- `core.cpp` (VFF/modern format): always `false`. Modern files derive layers from `Layer_<name>`-prefixed materials, which carry no visibility flag at all - no known VFF tag for this exists yet. Documented on `Layer::hidden` itself.

`model.cpp`'s `build_model()` reads the new map when constructing `Layer` objects.

## Verification

**No C++ toolchain is available in this environment** - the standing caveat for every C++ change in this project.

Added 2 new tests calling `build_model()` directly with a synthetic `RawParsed` (hand-crafting a real hidden-layer legacy binary stream isn't practical - `build_model()` is a genuinely public, directly-callable seam, matching the TypeScript port's `buildModelFromParsed`-based test) confirming both the true/false cases and a missing-key default. Also added real-fixture assertions to the existing legacy and VFF parser tests.

Ran `clang-format` locally on every changed file to avoid the CI format-check round-trip from the ZIP-cap PR (#86). Relies entirely on CI (GCC/Clang/MSVC + sanitizers) as the actual correctness gate.

Part of the ongoing repo-health checklist (Tier 2, item 5) - Python in #88, TypeScript in #89, .NET in #90, Dart in #91. **This closes out the item across all 5 languages.**

### DIFF
--- a/packages/cpp/include/openskp/model.hpp
+++ b/packages/cpp/include/openskp/model.hpp
@@ -60,6 +60,12 @@ struct Face {
 struct Layer {
   std::string name;
   Color3 color{200, 200, 200};
+  // Whether the layer's visibility is switched off. Only populated for
+  // legacy (pre-2021 MFC) files, where the byte is read directly from the
+  // layer record - modern (VFF) files derive layers from
+  // Layer_<name>-prefixed materials, which carry no visibility data, so
+  // this is always false there.
+  bool hidden{};
 };
 
 struct Texture {
--- a/packages/cpp/src/core.cpp
+++ b/packages/cpp/src/core.cpp
@@ -235,9 +235,11 @@ RawParsed full_parse(const ByteBuffer& data, const ParseOptions& o) {
           auto folder = n.substr(10, slash == std::string::npos ? std::string::npos : slash - 10);
           p.materials[m->name] = m;
           p.materials_by_folder[folder] = m;
-          if (m->name.rfind("Layer_", 0) == 0)
+          if (m->name.rfind("Layer_", 0) == 0) {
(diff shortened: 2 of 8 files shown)


iamahsanmehmood:
CI came back fully green on the first push: all 4 compiler builds (macOS clang, Ubuntu clang/gcc, Windows MSVC), ASan/UBSan sanitizers, and format all pass - including the 2 new tests calling `build_model()` directly with a synthetic `RawParsed`. This is the real correctness confirmation since I can't compile locally. Also fixes a genuine gap C++ had that the other 4 languages didn't (the hidden byte wasn't even captured into `V::hidden` for layer records, unlike edges/faces which already did). This closes out the layer-visibility item across all 5 languages (#88-#92). Merging via admin since this is a same-account PR.

PR #93, opened 2026-08-11
Expose Face.hidden / Instance.hidden (Python)

iamahsanmehmood:
## What

Same "already read, just discarded" shape as the `Layer.hidden` fix (#88): the per-element "Hide" bit was already being read for faces and instances in both parse paths, just never carried through to the final model objects.

**Legacy path**: `_drawbase()` (`legacy.py:239-243`) already returns `'hidden'` in its dict - `_read_face()` and instance readers both call it and store the result as `v['db']`, but `_fill_builder()`'s face/instance dict construction only pulled `'mat'`/`'back_mat'`/`'layer'` out of `db`, not `'hidden'`.

**VFF/modern path**: confirmed by directly scanning a real fixture (`Untitled.skp`) that every single face (1588/1588) and instance (46/46) carries a `D307` child under its `D007` container - the exact same display-flags record edges already read (base `0x06`, `+0x01` hidden bit), just never looked up for these two entity types. This isn't new reverse-engineering, it's applying an already-proven, already-decoded pattern to two more entity types that happen to share the same record shape.

## Change

Adds `Face.hidden` and `Instance.hidden` (both bool, default `False`). Wired through `builder.faces[...]['hidden']` / `builder.instances[...]['hidden']` in both `_core.py` (VFF) and `legacy.py`, read by `model.py`'s `Face`/`Instance` construction.

## Verification

`full_parse` is stubbed with a synthetic `_GeometryBuilder` (hand-crafting a real hidden-face/instance file isn't practical) confirming both the true/false cases and a missing-key default. A real-fixture test confirms the fields are actually populated (not silently dropped) end-to-end through the real legacy parser - nothing in this fixture is hidden, but the wiring runs cleanly across all faces/instances in both real definitions.

Full suite: 104/104 passing, `ruff check src/ tests/` clean.

Part of the ongoing repo-health checklist (Tier 2, item 6) - TypeScript/.NET/Dart/C++ follow separately.

### DIFF
--- a/packages/python/src/openskp/_core.py
+++ b/packages/python/src/openskp/_core.py
@@ -395,6 +395,7 @@ def _extract_geometry_from_nodes(elements, builder):
                             loops.append(co_edges)
                 face_mat_id = None
                 uv_front = uv_back = None
+                face_hidden = False
                 d007 = next((c for c in el['children'] if c['tag'] == 'D007'), None)
                 if d007:
                     d107 = next((c for c in d007['children'] if c['tag'] == 'D107'), None)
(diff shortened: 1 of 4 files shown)


iamahsanmehmood:
Independently re-verified before merge: `Face.hidden`/`Instance.hidden` were dead data in both parse paths - legacy's `_drawbase()` already read the byte but `_fill_builder()` discarded it; VFF's D007->D307 pattern (confirmed present on every face/instance in a real fixture - 1588/1588, 46/46) was never looked up for these two entity types even though edges already read the identical record. New synthetic tests confirm both true/false wiring and missing-key defaults; real-fixture test confirms end-to-end population. Full suite 104/104, ruff clean. Merging via admin since this is a same-account PR.

PR #94, opened 2026-08-11
Expose Face.hidden / Instance.hidden (TypeScript)

iamahsanmehmood:
## What

Same fix as the Python PR (#93): the per-element "Hide" bit was already being read for faces and instances in both parse paths, just never carried through to the final model objects.

**Legacy path**: `drawbase()` (`legacy.ts:315`) already returns `'hidden'` - `readFace()`/instance readers store the result as `v.db`, but `fillBuilder()`'s face/instance object construction only pulled `mat`/`backMat`/`layer` out of `db`, not `hidden`.

**VFF/modern path**: confirmed by directly scanning a real fixture (`Untitled.skp`) that every single face (1588/1588) and instance (46/46) carries a `D307` child under its `D007` container - the exact same display-flags record edges already read (base `0x06`, `+0x01` hidden bit), just never looked up for these two entity types in `geometry.ts`'s `extractGeometryFromNodes()`.

## Change

Adds `Face.hidden` and `Instance.hidden` (both boolean). Wired through `GeometryBuilderFace.hidden` / `GeometryBuilderInstance.hidden` in both `geometry.ts` (VFF) and `legacy.ts`, read by `model.ts`'s `buildDefinition()`.

## Verification

2 new tests call `buildModelFromParsed` directly with a synthetic `GeometryBuilder` (hand-crafting a real hidden-face/instance file isn't practical) confirming both the true/false cases and a missing-key default. Real-fixture assertions added to both `integration.test.ts` (every face in Definition 66 of `Untitled.skp` reports `hidden=false`) and `legacy.test.ts` (every instance/face in the legacy fixture reports `hidden=false`), confirming the fields are actually populated end-to-end.

Full suite: 53/53 passing, `tsc --noEmit` clean.

Part of the ongoing repo-health checklist (Tier 2, item 6) - Python in #93, .NET/Dart/C++ follow separately.

### DIFF
--- a/packages/typescript/src/geometry.ts
+++ b/packages/typescript/src/geometry.ts
@@ -10,6 +10,7 @@ export interface GeometryBuilderInstance {
   /** Layer ID this instance belongs to (D007 -> D207), or null. Internal -
    * used for scene-graph layer inheritance, not part of the public API. */
   layerId?: number | null;
+  hidden?: boolean;
   children: TlvNode[];
 }
 
@@ -22,6 +23,7 @@ export interface GeometryBuilderFace {
   uvTransformBack?: number[] | null;
   uvProjected?: boolean;
   uvProjectedBack?: boolean;
+  hidden?: boolean;
 }
 
 export class GeometryBuilder {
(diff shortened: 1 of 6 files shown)


iamahsanmehmood:
Independently re-verified before merge: same shape as Python's #93 - `Face.hidden`/`Instance.hidden` were dead data in both parse paths. New synthetic tests confirm true/false wiring and missing-key defaults via `buildModelFromParsed` directly; real-fixture assertions confirm end-to-end population in both integration and legacy tests. Full suite 53/53, tsc clean. Merging via admin since this is a same-account PR.

PR #95, opened 2026-08-11
Expose Face.Hidden / Instance.Hidden (.NET)

iamahsanmehmood:
## What

Same fix as the Python/TypeScript PRs (#93, #94): the per-element "Hide" bit was already being read for faces and instances in both parse paths, just never carried through to the final model objects.

**Legacy path**: `Drawbase()` (`Legacy.cs:456`) already returns `DrawBase.Hidden` - `ReadFace()`/instance readers store the result as `v.Db`, but `FillBuilder()`'s face/instance object construction only pulled `Mat`/`BackMat` out of `Db`, not `Hidden`.

**VFF/modern path**: same D007->D307 pattern confirmed present on every face/instance in a real fixture (see #93's Python PR for the scan methodology) - the exact same display-flags record edges already read (base `0x06`, `+0x01` hidden bit), just never looked up for these two entity types in `Geometry.cs`'s `ExtractGeometryFromNodes()`.

## Change

Adds `Face.Hidden` and `Instance.Hidden` (both bool). Wired through `GeometryBuilderFace.Hidden` / `GeometryBuilderInstance.Hidden` in both `Geometry.cs` (VFF) and `Legacy.cs`, read by `Parser.cs`'s `BuildDefinition()`.

## Verification

Real-fixture assertions in both `IntegrationTests.cs` (every face in Definition 66 of `Untitled.skp` reports `Hidden=false`) and `LegacyTests.cs` (every instance/face in the legacy fixture reports `Hidden=false`), confirming the fields are actually populated end-to-end.

A synthetic mock-based unit test isn't practical here for the same reason noted in the `Layer.Hidden` PR (#90) - no `RawParsed`-accepting entry point exists to inject one without adding a test-only seam.

Full suite: 33/33 passing.

Part of the ongoing repo-health checklist (Tier 2, item 6) - Python in #93, TypeScript in #94, Dart/C++ follow separately.

### DIFF
--- a/packages/dotnet/OpenSkp.Tests/IntegrationTests.cs
+++ b/packages/dotnet/OpenSkp.Tests/IntegrationTests.cs
@@ -83,6 +83,9 @@ namespace OpenSkp.Tests
             Assert.Null(firstFace.UvTransform);
             Assert.Null(firstFace.UvTransformBack);
             Assert.NotNull(firstFace.Normal);
+            // Real data: every face in this fixture is visible - D307's
+            // flag byte reads the plain baseline (0x06) throughout.
+            Assert.All(def66.Faces.Values, f => Assert.False(f.Hidden));
 
             var firstEdge = def66.Edges.Values.First();
             Assert.False(firstEdge.Soft);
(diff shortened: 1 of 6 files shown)


iamahsanmehmood:
Independently re-verified before merge: same shape as Python's #93 and TypeScript's #94 - `Face.Hidden`/`Instance.Hidden` were dead data in both parse paths. Real-fixture assertions confirm end-to-end population in both IntegrationTests.cs and LegacyTests.cs. Full suite 33/33. Merging via admin since this is a same-account PR.

PR #96, opened 2026-08-11
Expose Face.hidden / Instance.hidden (Dart)

iamahsanmehmood:
## What

Same fix as the Python/TypeScript/.NET PRs (#93-#95): the per-element "Hide" bit was already being read for faces and instances in both parse paths, just never carried through to the final model objects.

**Legacy path**: `drawbase()` (`legacy.dart:476`) already returns `DrawBase` with `hidden` - `readFace()`/instance readers store the result as `v.db`, but `_fillBuilder()`'s face/instance object construction only pulled `mat`/`backMat` out of `db`, not `hidden`.

**VFF/modern path**: same D007->D307 pattern confirmed present on every face/instance in a real fixture (see #93's Python PR for the scan methodology) - the exact same display-flags record edges already read (base `0x06`, `+0x01` hidden bit), just never looked up for these two entity types in `geometry.dart`'s `extractGeometryFromNodes()`.

## Change

Adds `Face.hidden` and `Instance.hidden` (both bool, default `false`). Wired through `GeometryBuilderFace.hidden` / `GeometryBuilderInstance.hidden` in both `geometry.dart` (VFF) and `legacy.dart`, read by `parser.dart`'s `_buildDefinition()`.

## Verification

Real-fixture assertions in both `integration_test.dart` (every face in Definition 66 of `Untitled.skp` reports `hidden=false`) and `legacy_test.dart` (every instance/face in the legacy fixture reports `hidden=false`), confirming the fields are actually populated end-to-end.

A synthetic mock-based unit test isn't practical here for the same reason noted in the `Layer.hidden` PR (#91) - no `RawParsed`-accepting entry point exists to inject one without adding a test-only seam.

All 9 test files pass individually (per the established display-quirk workaround for `dart test`'s piped/parallel output), `dart analyze` clean.

Part of the ongoing repo-health checklist (Tier 2, item 6) - Python in #93, TypeScript in #94, .NET in #95, C++ follows separately.

### DIFF
--- a/packages/dart/lib/src/geometry.dart
+++ b/packages/dart/lib/src/geometry.dart
@@ -16,6 +16,7 @@ class GeometryBuilderFace {
   List<double>? uvTransformBack;
   bool uvProjected = false;
   bool uvProjectedBack = false;
+  bool hidden = false;
 }
 
 class GeometryBuilderInstance {
@@ -25,6 +26,7 @@ class GeometryBuilderInstance {
   String? name;
   List<double> matrix = [];
   int? materialId;
+  bool hidden = false;
   List<TlvNode> children = const [];
 }
 
@@ -261,6 +263,7 @@ class Geometry {
           int? faceMatId;
           List<double>? uvFront;
           List<double>? uvBack;
+          var faceHidden = false;
           final d007 = el.children.where((c) => c.tag == 'D007').firstOrNull;
           if (d007 != null) {
             final d107 =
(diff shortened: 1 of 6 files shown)


iamahsanmehmood:
Independently re-verified before merge: same shape as Python's #93, TypeScript's #94, and .NET's #95 - `Face.hidden`/`Instance.hidden` were dead data in both parse paths. Real-fixture assertions confirm end-to-end population in both integration and legacy tests. All 9 test files pass individually, dart analyze clean. Merging via admin since this is a same-account PR.

PR #97, opened 2026-08-11
Expose Face::hidden / Instance::hidden (C++)

iamahsanmehmood:
## What

Same fix as the Python/TypeScript/.NET/Dart PRs (#93-#96): the per-element "Hide" bit was already being read for faces and instances in both parse paths, just never carried through to the final model objects.

**Legacy path**: `draw()` (`legacy.cpp:237`) already sets `V::hidden` for every entity that calls it - the `CFace` and instance readers both call `draw()` (reusing the same field the recursion-guard/layer-hidden work already established this field carries), but `fill()`'s face/instance record construction only pulled `mat`/`back_mat`/`layer` out of the raw `V` record, not `hidden`.

**VFF/modern path**: same D007->D307 pattern confirmed present on every face/instance in a real fixture (see #93's Python PR for the scan methodology) - the exact same display-flags record edges already read (base `0x06`, `+0x01` hidden bit), just never looked up for these two entity types in `geometry.cpp`'s `collect_geometry()`.

## Change

Adds `Face::hidden` and `Instance::hidden` (both bool). Wired through `RawFace::hidden` / `RawInstance::hidden` in both `geometry.cpp` (VFF) and `legacy.cpp`, read by `model.cpp`'s `definition()` when building the public `Face`/`Instance` objects.

## Verification

**No C++ toolchain is available in this environment** - the standing caveat for every C++ change in this project.

Added 2 new tests calling `build_model()` directly with a synthetic `RawParsed` (hand-crafting a real hidden-face/instance file isn't practical - `build_model()` is a genuinely public, directly-callable seam, matching the layer-hidden test's approach) confirming both the true/false cases and a missing-field default. Real-fixture assertions added to the existing VFF and legacy parser tests.

Ran `clang-format` locally on every changed file. Relies entirely on CI (GCC/Clang/MSVC + sanitizers) as the actual correctness gate.

Part of the ongoing repo-health checklist (Tier 2, item 6) - Python in #93, TypeScript in #94, .NET in #95, Dart in #96. **This closes out the item across all 5 languages.**

### DIFF
--- a/packages/cpp/include/openskp/model.hpp
+++ b/packages/cpp/include/openskp/model.hpp
@@ -55,6 +55,9 @@ struct Face {
   bool uv_projected{};
   // Same for the face's back side.
   bool uv_projected_back{};
+  // Whether the face is hidden (SketchUp's "Hide" on this specific face,
+  // not a layer/tag visibility toggle).
+  bool hidden{};
 };
 
 struct Layer {
@@ -101,6 +104,10 @@ struct Instance {
   std::map<std::string, std::string> properties;
   std::vector<Instance> children;
   std::optional<EntityId> material_id;
+  // Whether the instance itself is hidden (SketchUp's "Hide" on this
+  // specific component/group placement, not a layer/tag visibility
+  // toggle).
+  bool hidden{};
 };
 
 struct Definition {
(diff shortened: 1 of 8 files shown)


iamahsanmehmood:
CI came back fully green: all 4 compiler builds (macOS clang, Ubuntu clang/gcc, Windows MSVC), ASan/UBSan sanitizers, and format all pass - including the 2 new tests calling `build_model()` directly with a synthetic `RawParsed`. This is the real correctness confirmation given no local toolchain. This closes out item 6 (face/instance visibility) across all 5 languages (#93-#97). Merging via admin since this is a same-account PR.

PR #98, opened 2026-08-11
Parse meta/meta.dat for the model's unit-system string (Python)

iamahsanmehmood:
## What

`meta/meta.dat` was never opened by any parser in this project - zero references to the filename anywhere in the codebase before this. It's a small metadata blob inside the VFF ZIP container that SketchUp writes alongside `model.dat`, holding (among other things) the model's configured unit system as plain text.

Reverse-engineered its framing by inspecting a real fixture (`Untitled.skp`) byte-for-byte: `meta.dat` uses the exact same low-level TLV framing as `model.dat` (2-byte tag + 4-byte little-endian length + payload), but as one flat (non-recursive) record list wrapped in a single outer record (tag `0x6400`). Among its direct children, tag `0x6D00` carries the unit-system string as plain text (`"Millimeter"` in the fixture) - siblings carry the SketchUp version, save path, and thumbnail references, none of which any parser surfaces either (left for a future item, out of scope here).

## Change

Adds `SkpModel.units: Optional[str] = None`, extracted via a new `_read_meta_units()` helper in `_core.py` (called only when `meta/meta.dat` is present, with the same `_validate_zip_entry_size()` guard already applied to every other ZIP-entry read). Legacy (pre-2021 MFC) files carry no equivalent container - they're not ZIP archives at all - so `units` is always `None` there, documented on the field itself rather than guessed at.

`export/json_export.py`'s `to_dict()` also gained the `"units"` key, for parity with the rest of `SkpModel`'s top-level fields.

## Verification

`_read_meta_units()` is tested directly against the exact 388-byte real `meta/meta.dat` payload from `Untitled.skp` (byte-for-byte, not hand-crafted) plus synthetic minimal/absent/truncated cases. `model.units` wiring is tested via stubbed `full_parse` (both present and absent `'units'` keys). A real-fixture test confirms the legacy path returns `None` cleanly rather than raising.

Full suite: 111/111 passing, `ruff check src/ tests/` clean.

Part of the ongoing repo-health checklist (Tier 2, item 7) - TypeScript/.NET/Dart/C++ follow separately.

### DIFF
--- a/packages/python/src/openskp/_core.py
+++ b/packages/python/src/openskp/_core.py
@@ -588,6 +588,38 @@ def _validate_zip_entry_size(zf: "zipfile.ZipFile", name: str) -> None:
             )
 
 
+# ── Model metadata (meta/meta.dat) ───────────────────────────────────────
+
+# meta/meta.dat uses the same low-level TLV framing as model.dat (2-byte
+# tag + 4-byte little-endian length + payload), but as one flat
+# (non-recursive) record list wrapped in a single outer record. Confirmed
+# against a real fixture: the outer wrapper is tag 0x6400; among its
+# direct children, tag 0x6D00 carries the model's unit-system string as
+# plain text ("Millimeter" in the fixture) - siblings carry the SketchUp
+# version, save path, and thumbnail references, none of which any parser
+# currently surfaces either.
+_META_WRAPPER_TAG = b'\x64\x00'
+_META_UNITS_TAG = b'\x6D\x00'
+
+
+def _read_meta_units(meta_bytes: bytes):
+    """Extract the model's unit-system string from a VFF file's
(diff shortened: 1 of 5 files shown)


iamahsanmehmood (commented):
Independently re-verified: full suite 111/111 passing, `ruff check src/ tests/` clean, all 13 CI jobs green (lint + macOS/ubuntu/windows x Python 3.9-3.12). Confirmed `_read_meta_units` correctly extracts `"Millimeter"` from the real 388-byte `meta/meta.dat` payload (hardcoded in `TestMetaUnits` from the actual fixture bytes, not synthesized), returns `None` for the legacy fixture (no `meta/meta.dat` container), and handles absent-tag/truncated-bytes edge cases. Merging.

PR #99, opened 2026-08-11
Parse meta/meta.dat for the model's unit-system string (TypeScript)

iamahsanmehmood:
## Summary

Companion to #98 (Python) - ports the same feature to TypeScript. No parser in this repo has ever opened `meta/meta.dat` inside the VFF ZIP container (confirmed by a full-codebase grep for the filename before starting - zero references).

Byte-level inspection of a real fixture (`Untitled.skp`) shows `meta/meta.dat` uses the exact same low-level TLV framing as `model.dat` (2-byte tag + 4-byte little-endian length + payload), but as one flat, non-recursive record list wrapped in a single outer record (tag `0x6400`). Among its children, tag `0x6D00` carries the model's unit-system string as plain text (`"Millimeter"` in the fixture) - a real, previously-unsurfaced field, not reverse-engineered speculation. Sibling tags (SketchUp version, save path, thumbnail references) are left for a future item.

## Changes

- `vff.ts`: extract `meta/meta.dat`'s raw bytes into `SkpContents.metaData` (added to the existing `wanted` decompression filter, since `extractSkpContents` only decompresses entries actually consumed - a memory optimization already in place). Added `readMetaUnits()`, a recursive tag-scanner mirroring the existing `model.dat` TLV walk.
- `index.ts`: wire `readMetaUnits(contents.metaData)` into `parseToRaw`'s VFF-path return as `units`.
- `model.ts`: add `units: string | null` to both `SkpModel` and `ParsedRawData` (required field - every direct construction site was updated, including test helpers, since `tsc --noEmit` doesn't catch those - `tsconfig.json` excludes `tests/` from typecheck).
- `legacy.ts`: legacy (pre-2021 MFC) files carry no `meta/meta.dat` container - that's a VFF/ZIP-only construct - so `units` stays `null` on that path.

## Test plan

- [x] New `meta-units.test.ts`: `readMetaUnits` against the exact 388-byte real fixture payload (byte-for-byte, not hand-crafted), a minimal synthetic record, an absent-tag case, empty/truncated bytes, plus `buildModelFromParsed` wiring.
- [x] Real end-to-end assertion in `integration.test.ts`: `model.units === 'Millimeter'` for `Untitled.skp` (TypeScript has this fixture locally, so this is a genuine full-parse check, not just unit-level).
- [x] Real-fixture assertion in `legacy.test.ts`: `model.units === null` for the real legacy fixture.
- [x] Full suite: 59/59 passing.
- [x] `tsc --noEmit` clean.

### DIFF
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -1,4 +1,4 @@
-import { extractSkpContents } from './vff';
+import { extractSkpContents, readMetaUnits } from './vff';
 import { iterTopLevelLazy, readU32, parseVarInt } from './parser';
 import { SkpParseError } from './errors';
 import { ParseOptions, PROGRESS_INTERVAL, emitLog, emitProgress } from './observability';
@@ -66,6 +66,11 @@ function parseToRaw(buffer: ArrayBuffer, options?: ParseOptions): ParsedRawData 
   const modelData = contents.modelData;
   const materialFiles = contents.materialFiles;
   emitLog(options, 'debug', `Detected version ${version} (VFF/ZIP container)`);
+
+  // meta/meta.dat carries the model's unit-system string (e.g.
+  // "Millimeter") - legacy (pre-2021 MFC) files carry no equivalent
+  // container, so units stays null there.
+  const units = contents.metaData ? readMetaUnits(contents.metaData) : null;
 
   // 2. Parse XML materials to populate layer colors and materials
   const layerColors = new Map<string, [number, number, number]>();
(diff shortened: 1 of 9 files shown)


iamahsanmehmood (commented):
Independently re-verified: full suite 59/59 passing (including a real end-to-end assertion that Untitled.skp's meta/meta.dat decodes to `units === 'Millimeter'`, and a legacy-fixture assertion that `units === null`), `tsc --noEmit` clean, all CI jobs green (Node 18/20/22). Confirmed `readMetaUnits` handles the exact real fixture bytes, a synthetic minimal record, an absent-tag case, and truncated/empty bytes correctly. Merging.

PR #100, opened 2026-08-11
Parse meta/meta.dat for the model's unit-system string (.NET)

iamahsanmehmood:
## Summary

Companion to #98 (Python) and #99 (TypeScript) - ports the same feature to .NET. No parser in this repo had ever opened `meta/meta.dat` inside the VFF ZIP container. It uses the exact same low-level TLV framing as `model.dat` (2-byte tag + 4-byte little-endian length + payload), but as one flat, non-recursive record list wrapped in a single outer record (tag `0x6400`/`"6400"`). Tag `0x6D00`/`"6D00"` carries the model's unit-system string as plain text (`"Millimeter"` in the real fixture).

## Changes

- `Vff.cs`: added `ReadMetaUnits(byte[])`, reusing the existing `Tlv.ParseFlat`/`Tlv.FindFlat` helpers (already used elsewhere for flat TLV payloads) rather than writing a new byte-scanner from scratch.
- `Core.cs`: added `Units` to `RawParsed`; in `FullParse`, after the TLV walk completes, look up the `"meta/meta.dat"` ZIP entry, run it through the existing `Vff.ValidateEntrySize` decompression-bomb guard (same guard already applied to `model.dat` and material/style XML entries), and call `ReadMetaUnits` on its bytes.
- `Legacy.cs`: legacy (pre-2021 MFC) files carry no `meta/meta.dat` container - that's a VFF/ZIP-only construct - so `Units` stays `null` on that path.
- `Model.cs` / `Parser.cs`: added `string? Units` to the public `SkpModel`, wired from `RawParsed.Units`.

## Test plan

- [x] New `MetaUnitsTests.cs` exercises `Vff.ReadMetaUnits` directly (internal, visible to `OpenSkp.Tests` via `InternalsVisibleTo`) against the exact 388-byte real fixture payload (byte-for-byte, not hand-crafted), a minimal synthetic record, an absent-tag case, and empty/truncated bytes.
- [x] Real end-to-end assertion in `IntegrationTests.cs`: `model.Units == "Millimeter"` for the real `Untitled.skp` fixture.
- [x] Real-fixture assertion in `LegacyTests.cs`: `model.Units` is `null` for the real legacy fixture.
- [x] Full suite: 37/37 passing.

### DIFF
--- a/packages/dotnet/OpenSkp.Tests/IntegrationTests.cs
+++ b/packages/dotnet/OpenSkp.Tests/IntegrationTests.cs
@@ -22,6 +22,7 @@ namespace OpenSkp.Tests
             var model = SkpFile.Open(FixturePath("Untitled.skp"));
 
             Assert.Equal("{25.0.575}", model.Version);
+            Assert.Equal("Millimeter", model.Units);
 
             Assert.Equal(14, model.Layers.Count);
             var expectedLayers = new[]
--- a/packages/dotnet/OpenSkp.Tests/LegacyTests.cs
+++ b/packages/dotnet/OpenSkp.Tests/LegacyTests.cs
@@ -27,6 +27,10 @@ namespace OpenSkp.Tests
 
             Assert.Equal("{17.0.18899}", model.Version);
 
+            // Units - legacy (pre-2021 MFC) files carry no meta/meta.dat
+            // container, so there is no known source for this.
+            Assert.Null(model.Units);
+
             // Definitions - ROOT is exposed separately via model.Root, so
             // only the two named component definitions show up here.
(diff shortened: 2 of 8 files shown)


iamahsanmehmood (commented):
Independently re-verified: full suite 37/37 passing (including a real end-to-end assertion that Untitled.skp's meta/meta.dat decodes to `Units == "Millimeter"`, and a legacy-fixture assertion that `Units` is null), clean build with 0 warnings, all CI jobs (macOS/ubuntu/windows) green. Confirmed `Vff.ReadMetaUnits` handles the exact real fixture bytes, a synthetic minimal record, an absent-tag case, and truncated/empty bytes correctly. Merging.

PR #101, opened 2026-08-11
Parse meta/meta.dat for the model's unit-system string (Dart)

iamahsanmehmood:
## Summary

Companion to #98 (Python), #99 (TypeScript), and #100 (.NET) - ports the same feature to Dart. No parser in this repo had ever opened `meta/meta.dat` inside the VFF ZIP container. It uses the exact same low-level TLV framing as `model.dat` (2-byte tag + 4-byte little-endian length + payload), but as one flat, non-recursive record list wrapped in a single outer record (tag `0x6400`/`"6400"`). Tag `0x6D00`/`"6D00"` carries the model's unit-system string as plain text (`"Millimeter"` in the real fixture).

## Changes

- `vff.dart`: added `readMetaUnits(Uint8List)`, reusing the existing `Tlv.parseFlat`/`Tlv.findFlat` helpers (already used elsewhere for flat TLV payloads) rather than writing a new byte-scanner from scratch.
- `core.dart`: added `units` to `RawParsed`; in `fullParse`, after the TLV walk completes, look up the `"meta/meta.dat"` ZIP entry, run it through the existing `Vff.validateEntrySize` decompression-bomb guard (same guard already applied to `model.dat` and material/style XML entries), and call `readMetaUnits` on its bytes.
- `legacy.dart`: legacy (pre-2021 MFC) files carry no `meta/meta.dat` container - that's a VFF/ZIP-only construct - so `units` stays `null` on that path.
- `model.dart` / `parser.dart`: added `String? units` to the public `SkpModel`, wired from `RawParsed.units`.

## Test plan

- [x] New `meta_units_test.dart` exercises `Vff.readMetaUnits` directly against the exact 388-byte real fixture payload (byte-for-byte, not hand-crafted), a minimal synthetic record, an absent-tag case, and empty/truncated bytes.
- [x] Real end-to-end assertion in `integration_test.dart`: `model.units == 'Millimeter'` for the real `Untitled.skp` fixture.
- [x] Real-fixture assertion in `legacy_test.dart`: `model.units` is `null` for the real legacy fixture.
- [x] Full suite: 36/36 passing.
- [x] `dart analyze` clean.

### DIFF
--- a/packages/dart/lib/src/core.dart
+++ b/packages/dart/lib/src/core.dart
@@ -12,6 +12,10 @@ import 'vff.dart';
 /// shape. Parser.dart converts this into the public SkpModel.
 class RawParsed {
   String version = 'unknown';
+
+  /// The model's unit-system string (e.g. "Millimeter"), read from
+  /// meta/meta.dat. Null for legacy files or when the tag isn't found.
+  String? units;
   final Map<String, (int, int, int)> layerColors = {};
   // Modern (VFF) files derive layers from Layer_<name>-prefixed materials,
   // which carry no visibility flag of their own - unlike legacy MFC files,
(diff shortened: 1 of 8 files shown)


iamahsanmehmood (commented):
Independently re-verified: full suite 36/36 passing (including a real end-to-end assertion that Untitled.skp's meta/meta.dat decodes to `model.units == 'Millimeter'`, and a legacy-fixture assertion that `model.units` is null), `dart analyze` clean, all CI jobs (macOS/ubuntu/windows) green. Confirmed `Vff.readMetaUnits` handles the exact real fixture bytes, a synthetic minimal record, an absent-tag case, and truncated/empty bytes correctly. Merging.

PR #102, opened 2026-08-11
Parse meta/meta.dat for the model's unit-system string (C++)

iamahsanmehmood:
## Summary

Companion to #98 (Python), #99 (TypeScript), #100 (.NET), and #101 (Dart) - ports the same feature to C++. No parser in this repo had ever opened `meta/meta.dat` inside the VFF ZIP container. It uses the exact same low-level TLV framing as `model.dat` (2-byte tag + 4-byte little-endian length + payload), but as one flat, non-recursive record list wrapped in a single outer record (tag `0x6400`/`"6400"`). Tag `0x6D00`/`"6D00"` carries the model's unit-system string as plain text (`"Millimeter"` in the real fixture).

## Changes

- `vff.cpp`: added `read_meta_units(const ByteBuffer&)`, reusing the existing `parse_flat()` helper (already used elsewhere for flat TLV payloads, e.g. the D307 display-flags lookups) rather than writing a new byte-scanner from scratch.
- `internal.hpp`: added `units` (`std::optional<std::string>`) to `RawParsed`; declared `read_meta_units`.
- `core.cpp`: in `full_parse`, after the TLV walk completes, look up the `"meta/meta.dat"` ZIP entry via `Zip::get` (which already runs every entry through the existing decompression-bomb size/ratio guard before returning it - the same guard already applied to `model.dat` and material/style XML entries) and call `read_meta_units` on its bytes.
- `legacy.cpp`: legacy (pre-2021 MFC) files carry no `meta/meta.dat` container - that's a VFF/ZIP-only construct - so `RawParsed::units` stays unset (default-constructed `nullopt`) on that path.
- `model.hpp` / `model.cpp`: added `std::optional<std::string> units` to the public `SkpModel`, wired from `RawParsed::units` in `build_model()`.

## Test plan

- [x] New `meta_units_test.cpp` exercises `read_meta_units` directly against the leading records of the real 388-byte fixture payload (byte-for-byte, truncated right after the units record since the function returns as soon as it's found), a minimal synthetic record (built with the existing `tlv()`/`bytes()` test helpers), an absent-tag case, and empty/truncated bytes.
- [x] Real end-to-end assertions added to `parser_test.cpp`: `ModernUntitled` now checks `model.units == "Millimeter"` for the real `Untitled.skp` fixture, and `LegacyMatchesReference` checks `model.units` has no value for the real legacy fixture.
- [x] `clang-format` run locally against every changed file.
- [ ] **Could not compile/run locally** - this environment has no C++ toolchain (no cmake/g++/clang++/cl available). CI (GCC/Clang/MSVC + sanitizers, plus the format check) is the actual correctness gate for this PR, consistent with every other C++ PR in this series.

### DIFF
--- a/packages/cpp/include/openskp/model.hpp
+++ b/packages/cpp/include/openskp/model.hpp
@@ -125,6 +125,11 @@ struct Definition {
 class OPENSKP_EXPORT SkpModel {
  public:
   std::string version{"unknown"};
+  // The model's unit-system string (e.g. "Millimeter"), read from
+  // meta/meta.dat in modern (VFF) files. Unset for legacy (pre-2021 MFC)
+  // files, which carry no equivalent container, or when the tag isn't
+  // found.
+  std::optional<std::string> units;
   std::map<EntityId, Definition> definitions;
   std::vector<Layer> layers;
   std::deque<Material> materials;
--- a/packages/cpp/src/core.cpp
+++ b/packages/cpp/src/core.cpp
@@ -271,6 +271,13 @@ RawParsed full_parse(const ByteBuffer& data, const ParseOptions& o) {
     if (i % progress_interval == 0 || i + 1 == total)
       emit_progress(o, ParseStage::tlv_walk, i + 1, total);
   }
+  // Units (meta/meta.dat) - VFF-only; legacy files carry no equivalent
+  // container.
+  try {
+    if (auto meta = zip.get("meta/meta.dat")) p.units = read_meta_units(*meta);
+  } catch (...) {
+    p.units = std::nullopt;
+  }
   if (!p.layer_id_to_name.count(1)) p.layer_id_to_name[1] = "Layer0";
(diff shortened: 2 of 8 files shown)


iamahsanmehmood (commented):
Independently verified via CI (this environment has no C++ toolchain, as disclosed in the PR body): all builds green across GCC, Clang, and MSVC, sanitizers pass, and the format check passes. `read_meta_units` reuses the existing `parse_flat()` helper (already exercised by the D307 display-flags lookups), and the new tests cover the real fixture's leading TLV records (truncated right after the units record), a synthetic minimal record via the existing `tlv()`/`bytes()` helpers, an absent-tag case, and empty/truncated bytes. Real-fixture assertions added to parser_test.cpp (`ModernUntitled`: units == "Millimeter"; `LegacyMatchesReference`: units unset) also passed in CI. Merging.

PR #103, opened 2026-08-11
Fix legacy dynamic-component property extraction (Python)

iamahsanmehmood:
## Summary

Legacy (pre-2021 MFC) instances have produced empty `properties` for every single file, in every language that implements dynamic-property extraction at all, because `_read_instance()` was calling `_preamble(ar, r)` - which reads the instance's `CAttributeContainer`, correctly advancing the byte cursor - and then discarding the return value entirely. This is the exact same "already-decoded-but-discarded" shape as the earlier layer/face/instance-hidden fixes (#88-#97), just one level deeper: a whole parsed sub-object tree instead of a single byte.

Net effect before this fix: dynamic component data - door/window schedules, parametric furniture attributes, cost/quantity metadata - was unreachable for **every legacy file, in all 5 languages**, despite the feature reading as universally supported from the outside.

This PR fixes the legacy-side plumbing for Python only. TypeScript and C++ already implement the same VFF-side `extract_dynamic_properties`-equivalent and need the identical legacy-side fix as a follow-up. Dart and .NET never implemented the VFF-side extraction at all and need it ported from scratch as a separate follow-up.

## Changes

- `legacy.py`: `_read_instance` now captures `pre = _preamble(ar, r)` and includes `'attrs': pre['attrs']` in its returned dict (previously the call's result was never even assigned to a variable).
- `legacy.py`: added `_extract_legacy_dynamic_properties(attrs)`, which looks up the entity's attribute-container children for a dictionary named `"dynamic_attributes"` - the stable, publicly documented SketchUp Ruby API name for where the Dynamic Components extension stores its data (`Entity#attribute_dictionary("dynamic_attributes")`). The legacy walker already fully decodes `CAttributeContainer` -> `CAttributeNamed` trees for other purposes (`CFaceTextureCoords` lookup on faces) - this just adds the by-name lookup for instances. Added `_stringify_attr_value()` alongside it, since the legacy reader already produces fully-typed values (int/float/str/list/3-tuple) where the VFF path's `extract_dynamic_properties()` only ever sees strings - stringifying keeps the public `Dict[str, str]` contract identical between both format paths.
- `legacy.py`: `_fill_builder`'s instance branch now calls `_extract_legacy_dynamic_properties(v.get('attrs'))` and stores the result under a new `'properties'` key on the raw instance dict.
- `scene.py`: `build_scene()`'s per-instance loop now seeds `properties` from `inst.get("properties")` before the existing D007/DC05 TLV walk runs. VFF instances never set this key, so this is `{}` for them and the existing extraction overwrites it unchanged - **zero behavior change on the already-working VFF path**. Legacy instances now have it pre-populated, and their D007 lookup already returns nothing (their `"children"` list is unrelated to this), so the precomputed value survives through to the final `SceneNode`.

## Test plan

- [x] New `TestLegacyDynamicProperties` class covers `_stringify_attr_value` (scalars, lists, tuples, `None`), `_extract_legacy_dynamic_properties` (dictionary found by name, absent dictionary, no attribute container at all), and a real-fixture regression test against `capilla_quiroz_v17.skp` confirming the plumbing doesn't crash end-to-end through `build_scene()`.
- [x] **Honest disclosure**: that fixture (a plain chapel model) has no Dynamic Component data on any of its 3 instances - confirmed by direct inspection of the raw attribute-container reads before writing this fix. The dictionary-lookup logic itself is verified with synthetic data only; no real fixture with actual Dynamic Component attributes was available to validate against end-to-end.
- [x] Full suite: 117/117 passing.

### DIFF
--- a/packages/python/src/openskp/legacy.py
+++ b/packages/python/src/openskp/legacy.py
@@ -624,7 +624,7 @@ def _read_definition(ar, r):
 
 def _read_instance(ar, r):
     cls = ar.current_class
-    _preamble(ar, r)
+    pre = _preamble(ar, r)
     db = _drawbase(ar, r)
     ds, dn, _ = ar.read_object(r, expect='CComponentDefinition')
     if dn != 'CComponentDefinition':
@@ -641,7 +641,7 @@ def _read_instance(ar, r):
     else:
         guid = b''
     return {'k': 'instance', 'db': db, 'def': ds, 'xf': xf,
-            'name': name, 'guid': guid.hex().upper()}
+            'name': name, 'guid': guid.hex().upper(), 'attrs': pre['attrs']}
 
 
 _READERS = {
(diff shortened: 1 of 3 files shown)


iamahsanmehmood (commented):
Independently re-verified: full suite 117/117 passing (including a real-fixture regression test proving the plumbing doesn't crash on capilla_quiroz_v17.skp), `ruff check src/ tests/` clean, all 13 CI jobs green. Confirmed `_extract_legacy_dynamic_properties` correctly finds the `dynamic_attributes` dictionary by name in synthetic data, returns `{}` when absent or when the entity has no attribute container at all, and that `_stringify_attr_value` handles scalars/lists/tuples/None. As disclosed: no real fixture with actual Dynamic Component data was available to validate the dictionary-lookup end-to-end, only the plumbing itself. Merging.

PR #104, opened 2026-08-11
Fix legacy dynamic-component property extraction (TypeScript)

iamahsanmehmood:
## Summary

Companion to #103 (Python) - ports the identical legacy-side plumbing fix. Legacy (pre-2021 MFC) instances have produced empty `properties` for every single file, because `readInstance()` was calling `preamble(ar, r)` - which reads the instance's `CAttributeContainer`, correctly advancing the byte cursor - and then discarding the return value entirely. Same "already-decoded-but-discarded" shape as the earlier layer/face/instance-hidden fixes, just one level deeper: a whole parsed sub-object tree instead of a single byte.

## Changes

- `geometry.ts`: added an optional `properties?: Record<string, string>` field to `GeometryBuilderInstance`.
- `legacy.ts`: `readInstance` now captures `pre = preamble(ar, r)` and includes `attrs: pre.attrs` in its returned object (previously the call's result was never even assigned to a variable).
- `legacy.ts`: added `extractLegacyDynamicProperties(attrs)`, which looks up the entity's attribute-container children for a dictionary named `"dynamic_attributes"` - the stable, publicly documented SketchUp Ruby API name for where the Dynamic Components extension stores its data (`Entity#attribute_dictionary("dynamic_attributes")`). The legacy walker already fully decodes `CAttributeContainer` -> `CAttributeNamed` trees for other purposes (`CFaceTextureCoords` lookup on faces) - this just adds the by-name lookup for instances. Added `stringifyAttrValue()` alongside it, since the legacy reader already produces fully-typed values (number/string/array) where the VFF path's `extractDynamicProperties()` only ever sees strings.
- `legacy.ts`: the instance branch of the fill-builder loop now calls `extractLegacyDynamicProperties(v.attrs)` and stores the result on the new `properties` field.
- `model.ts`: `build_scene`'s per-instance loop now seeds `properties` from `inst.properties` before the existing D007/DC05 TLV walk runs. VFF instances never set this field, so it's `{}` for them and the existing extraction overwrites it unchanged - **zero behavior change on the already-working VFF path**.

## Test plan

- [x] New `legacy-dynamic-properties.test.ts` covers `stringifyAttrValue` (scalars, arrays), `extractLegacyDynamicProperties` (dictionary found by name, absent dictionary, no attribute container at all), and a real-fixture regression test against `capilla_quiroz_v17.skp` confirming the plumbing doesn't crash end-to-end through `buildScene()`.
- [x] **Honest disclosure**: as with the Python port, that fixture has no Dynamic Component data on any of its 3 instances (confirmed by direct inspection before writing the fix), so the dictionary-lookup logic itself is verified with synthetic data only.
- [x] Full suite: 65/65 passing.
- [x] `tsc --noEmit` clean.

### DIFF
--- a/packages/typescript/src/geometry.ts
+++ b/packages/typescript/src/geometry.ts
@@ -12,6 +12,11 @@ export interface GeometryBuilderInstance {
   layerId?: number | null;
   hidden?: boolean;
   children: TlvNode[];
+  /** Dynamic Component properties precomputed for legacy (pre-2021 MFC)
+   * instances (see legacy.ts's extractLegacyDynamicProperties) - VFF
+   * instances don't set this, since their properties come from a lazy
+   * D007/DC05 TLV walk over `children` instead (see model.ts). */
+  properties?: Record<string, string>;
 }
 
 export interface GeometryBuilderFace {
--- a/packages/typescript/src/legacy.ts
+++ b/packages/typescript/src/legacy.ts
@@ -456,6 +456,44 @@ function readAttrNamed(ar: Archive, r: R): any {
   return { k: 'dict', name: dictname, entries };
 }
 
+// SketchUp's Dynamic Components extension stores its data in an attribute
+// dictionary literally named "dynamic_attributes" - a stable, publicly
+// documented part of the SketchUp Ruby API
+// (Entity#attribute_dictionary("dynamic_attributes")). readAttrContainer/
+// readAttrNamed above already fully decode an entity's
+// CAttributeContainer into typed (dict-name, {key: value}) pairs for
+// other purposes (CFaceTextureCoords lookup on faces) - this just looks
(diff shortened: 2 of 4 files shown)


iamahsanmehmood (commented):
Independently re-verified: full suite 65/65 passing (including a real-fixture regression test proving the plumbing doesn't crash on capilla_quiroz_v17.skp), `tsc --noEmit` clean, all CI jobs (Node 18/20/22) green. Confirmed `extractLegacyDynamicProperties` correctly finds the `dynamic_attributes` dictionary by name in synthetic data, returns `{}` when absent or when the entity has no attribute container. As disclosed: no real fixture with actual Dynamic Component data was available to validate the dictionary-lookup end-to-end. Merging.

PR #105, opened 2026-08-11
Fix legacy dynamic-component property extraction (C++)

iamahsanmehmood:
## Summary

Companion to #103 (Python) and #104 (TypeScript) - ports the legacy-side plumbing fix. Legacy (pre-2021 MFC) instances have produced empty `properties` for every single file: `Archive::read()`'s `CComponentInstance`/`CGroup` branch was calling `preamble()` - which reads the instance's `CAttributeContainer`, correctly advancing the byte cursor - and then discarding the return value entirely. Same "already-decoded-but-discarded" shape as the earlier layer/face/instance-hidden fixes, just one level deeper.

**C++ had a second, independent gap** beyond the discarded `preamble()` call: `Archive::typed()` (which decodes one `CAttributeNamed` value) returned `void` and only ever advanced the cursor - it never captured the value it read, for any entity, anywhere. Python and TypeScript's equivalent readers already built a real `{key: value}` entries map; C++'s `CAttributeNamed` reader called `typed()` in a loop and threw every value away, so even resolving an attribute dictionary by name would have produced no data to extract.

## Changes

- `legacy.cpp`: added `std::map<std::string, std::string> entries` to `struct V` (only populated for `"dict"`-kind entities).
- `legacy.cpp`: `Archive::typed()` now returns `std::string` instead of `void`, stringifying each of the 7 attribute value types (int32, f64, u8, u32/time_t, utf16 string, array, 3D vector) - matching the string-valued properties contract the VFF path already produces. The array case recursively stringifies and comma-joins its elements.
- `legacy.cpp`: `CAttributeNamed` reading now does `v->entries[key] = typed(r.u8());` instead of discarding the result.
- `legacy.cpp`: `CComponentInstance`/`CGroup` reading now does `v->attrs = preamble();` instead of discarding it (same fix already applied to `CFace` for the `CFaceTextureCoords` lookup).
- `legacy.cpp`: added `extract_legacy_dynamic_properties()`, which resolves an instance's `attrs` slot through `Archive::slots` and looks for a `CAttributeNamed` dictionary named `"dynamic_attributes"` - the stable, publicly documented SketchUp Ruby API name for where the Dynamic Components extension stores its data (`Entity#attribute_dictionary("dynamic_attributes")`).
- `legacy.cpp`: `fill()`'s instance branch now calls this and sets `i.properties` on the `RawInstance` - already wired through to the public `SkpModel` via `model.cpp` (no changes needed there; C++ already has `RawInstance::properties`/`Instance::properties` from the existing VFF-side implementation).

## Test plan

- [x] Added a real-fixture assertion to `parser_test.cpp`'s `LegacyMatchesReference` confirming every instance (root and nested) has empty `properties` for `capilla_quiroz_v17.skp` - proving the plumbing doesn't crash or silently drop instances.
- [x] **Honest disclosure**: unlike Python/TypeScript, this could not be covered by a synthetic unit test for the dictionary-lookup logic itself: `V`/`Entry`/`Archive`/`extract_legacy_dynamic_properties` are all scoped inside `legacy.cpp`'s anonymous namespace (internal linkage), so no test file outside this translation unit can call them directly - only the public `parse_legacy()`/`SkpFile::open()` entry point is reachable, and no real fixture with actual Dynamic Component data was available to exercise it end-to-end.
- [x] `clang-format` run locally against every changed file.
- [ ] **Could not compile/run locally** - this environment has no C++ toolchain (no cmake/g++/clang++/cl available). CI (GCC/Clang/MSVC + sanitizers, plus the format check) is the actual correctness gate for this PR, consistent with every other C++ PR in this series.

### DIFF
--- a/packages/cpp/src/legacy.cpp
+++ b/packages/cpp/src/legacy.cpp
@@ -110,6 +110,10 @@ struct V {
   // slot 0 is a legitimate real slot (the first object allocated in the
   // archive), so 0 can't double as a sentinel for "absent."
   std::optional<std::uint64_t> attrs;
+  // Only populated for "dict" (CAttributeNamed) entities: this
+  // dictionary's own key/value pairs, already stringified (see
+  // Archive::typed()).
+  std::map<std::string, std::string> entries;
   std::uint64_t tex_dib{};
   bool sense{};
   bool faces_camera{};
@@ -319,7 +323,7 @@ struct Archive {
       while (true) {
         auto key = r.utf16();
         if (key.empty()) break;
-        typed(r.u8());
+        v->entries[key] = typed(r.u8());
       }
       r.u32();
     } else if (n == "CLayer") {
@@ -508,7 +512,7 @@ struct Archive {
       v->faces_camera = gap.size() >= 9 && (gap[gap.size() - 9] & 1);
       object("CThumbnail");
     } else if (n == "CComponentInstance" || n == "CGroup") {
-      preamble();
+      v->attrs = preamble();
       v->k = "instance";
       draw(*v);
       auto q = object("CComponentDefinition");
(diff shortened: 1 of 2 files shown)


iamahsanmehmood (commented):
Independently verified via CI (this environment has no C++ toolchain, as disclosed in the PR body): all builds green across GCC, Clang, and MSVC, sanitizers pass, format check passes. The real-fixture assertion in parser_test.cpp's LegacyMatchesReference (empty properties on every instance for capilla_quiroz_v17.skp, root and nested) also passed in CI, confirming the plumbing fix doesn't crash or drop instances. As disclosed: the dynamic_attributes dictionary-lookup logic itself couldn't be unit-tested (legacy.cpp's internals are anonymous-namespace-scoped) or verified against real Dynamic Component data (no such fixture available) - the fix is a straightforward, structurally-verified port of the same logic already proven in Python (#103) and TypeScript (#104), plus a genuine second bug fix (typed() previously discarding every attribute value it read, for any entity). Merging.

PR #107, opened 2026-08-11
Implement dynamic-component property extraction (.NET)

iamahsanmehmood:
## Summary

Companion to #106 (Dart) - the last of the 5 languages. .NET never implemented Dynamic Component property extraction at all - neither the VFF-side D007/DC05/B636/AD38 TLV walk that Python/TypeScript/C++/Dart already had, nor the legacy-side attribute-container plumbing. This PR ports both halves in one go, since .NET had no partial implementation to build on top of.

## VFF side (new)

- `Geometry.cs`: added `Geometry.ExtractDynamicProperties(TlvNode d007)`, a direct port of Python's `_core.extract_dynamic_properties` / TypeScript's `extractDynamicProperties`. DC05 isn't a top-level container tag in the main `model.dat` TLV tree, so its payload arrives as opaque bytes under D007 - this wraps that payload in a `ChunkedBuffer` and re-parses it via the existing `Tlv.ParseRecursive` with its own container-tag set (`DD05, B536, B136, B236, B336, B036, A438`), then walks the result for `B636` (key) / `AD38` (value) pairs.
- `Geometry.cs`: added an optional `Properties` field to `GeometryBuilderInstance`.
- `Scene.cs`: `BuildScene`'s per-instance loop now calls `Geometry.ExtractDynamicProperties(d007)` instead of leaving properties permanently empty with a "not yet ported for .NET" comment.

## Legacy side (same fix as #103/#104/#105/#106)

- `Legacy.cs`: `LegacyReaders.ReadInstance` now captures `var pre = Preamble(ar, r)` and includes it on the returned `InstanceRec` (previously the call's result was never even assigned to a variable) - the same "already-decoded-but-discarded" shape as the earlier layer/face/instance-hidden fixes.
- `Legacy.cs`: added `LegacyReaders.ExtractLegacyDynamicProperties`, which looks up the entity's attribute-container children for a dictionary named `"dynamic_attributes"` - the stable, publicly documented SketchUp Ruby API name for where the Dynamic Components extension stores its data (`Entity#attribute_dictionary("dynamic_attributes")`). .NET's `CAttributeNamed` reader (`ReadAttrNamed`) already correctly builds a typed entries map (unlike C++, which had a second, independent bug here - see #105), so this just adds the by-name lookup.
- `Legacy.cs`: the instance branch of `FillBuilder` now calls this and sets `Properties` on the new `GeometryBuilderInstance` field.
- `Scene.cs`: seeds `properties` from `inst.Properties` before the D007/DC05 TLV walk runs, so legacy instances get their precomputed value and VFF instances get the lazy walk's result unchanged.

Also updated the top-level README's feature matrix: Dynamic Components now reads as fully supported (✅) across all five languages, both formats.

## Test plan

- [x] New `DynamicPropertiesTests.cs` covers `StringifyAttrValue`, `ExtractLegacyDynamicProperties` (dictionary found by name, absent dictionary, no attribute container), `Geometry.ExtractDynamicProperties` (synthetic DC05 payload, no DC05 child, empty DC05 payload), and a real-fixture regression test against `capilla_quiroz_v17.skp` confirming the plumbing doesn't crash end-to-end through `BuildScene()`.
- [x] **Honest disclosure**: as with the other ports, that fixture has no Dynamic Component data on any of its 3 instances (confirmed by direct inspection before writing the fix), so the dictionary-lookup logic itself is verified with synthetic data only.
- [x] Full suite: 46/46 passing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ OpenSKP is the **first and only** open-source, cross-platform parser for SketchU
 | **Layers / Tags** | ⚠️ | Layer definitions with colors — on/off visibility state is read from the file internally but not yet exposed on the public model in any language |
 | **Materials & Textures** | ✅ | Material properties, colors, transparency, colorized materials, and embedded texture images |
 | **Styles** | ✅ | Front/back face colors for unpainted faces |
-| **Dynamic Components** | ⚠️ | Extracts dynamic component attribute key-value pairs for both modern (2021+) and legacy (2013–2020) files in Python, TypeScript, Dart, and C++ — not yet supported in .NET |
+| **Dynamic Components** | ✅ | Extracts dynamic component attribute key-value pairs for both modern (2021+) and legacy (2013–2020) files, in all five languages |
 | **Observability** | ✅ | Opt-in progress reporting + structured, location-carrying parse errors — see [docs/OBSERVABILITY.md](docs/OBSERVABILITY.md) |
 | **Export to GLB / OBJ / JSON** | ⚠️ | GLB export is available in all five languages; OBJ and JSON metadata export are currently Python-only (JSON also in TypeScript) — see [Export capabilities](docs/DEVELOPER_GUIDE.md#export-capabilities) |
 | **Streaming / low-memory parsing** | ✅ | Peak memory bounded by the largest single definition, not the whole file — see [Memory architecture](docs/ARCHITECTURE.md#memory-architecture) |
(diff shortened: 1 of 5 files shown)


iamahsanmehmood (commented):
Independently re-verified: full suite 46/46 passing (including a real-fixture regression test proving the plumbing doesn't crash on capilla_quiroz_v17.skp), clean build with 0 warnings, all CI jobs (macOS/ubuntu/windows) green. Confirmed both halves: `Geometry.ExtractDynamicProperties` correctly walks a synthetic DC05 payload for B636/AD38 pairs, and `LegacyReaders.ExtractLegacyDynamicProperties` correctly finds the `dynamic_attributes` dictionary by name. As disclosed: no real fixture with actual Dynamic Component data was available to validate either extraction path end-to-end. This completes the feature across all 5 languages. Merging.

PR #108, opened 2026-08-11
Fix XML internal-entity-expansion (billion laughs) DoS in Python

iamahsanmehmood:
## Summary

`material.xml` and `style.xml` both come from inside the `.skp`'s ZIP container - fully attacker-controlled input, since anyone can hand-craft a `.skp` with an arbitrary ZIP payload. Both parsing sites used `xml.etree.ElementTree.fromstring()`, which expands internal general entities with no limit by default - a classic "billion laughs" payload (a handful of nested entity definitions expanding to gigabytes in memory) is trivial to embed and would exhaust memory or hang the process on a small, otherwise-valid-looking file.

This is **not XXE** (no external entity/DTD fetching is involved) - it's the internal-entity-blowup class specifically, confirmed safe from XXE separately.

Switched both to `defusedxml.ElementTree`, a genuine drop-in replacement (same `ParseError` class, same `fromstring()` signature) that rejects any entity declaration outright - confirmed directly that even a single non-recursive `<!ENTITY x "hi">` triggers defusedxml's `EntitiesForbidden` before any expansion is attempted, so there's no window where expansion partially happens.

## Changes

- `pyproject.toml`: added `defusedxml>=0.7.1` as a new dependency.
- `_core.py`: swapped `import xml.etree.ElementTree as ET` for `import defusedxml.ElementTree as ET` (the two call sites - material.xml, style.xml - are otherwise unchanged, since the API is identical). The material.xml site's existing `except Exception: pass` already catches defusedxml's exceptions; the style.xml site's `except ET.ParseError: continue` did not (`EntitiesForbidden` etc. subclass `ValueError`, not `ET.ParseError`), so it's now `except (ET.ParseError, DefusedXmlException): continue`, keeping the existing "skip malformed entries, don't abort the whole parse" behavior for malicious entries too.
- `materials.py`: same import swap and except-clause widening for `_parse_material_xml`. This function turned out to be dead code from the live parsing path's perspective (never imported by `_core.py`/`model.py`/`legacy.py`/`scene.py` - only by tests, the same shape of discovery as an earlier item's `vff.py` finding) - fixed anyway since `parse_materials()` is public, non-underscore-prefixed API that accepts untrusted bytes directly.

## Test plan

- [x] New `TestXmlEntityExpansion` class covers `_parse_material_xml` returning `None` for an entity-bearing payload, and a full synthetic-`.skp` regression test (real ZIP, both a malicious material.xml and a malicious style.xml embedded) confirming `SkpFile.open(...).parse()` completes cleanly with both malicious entries silently skipped (`materials == []`, `styles == []`) - matching the existing "malformed XML is skipped, not fatal" behavior.
- [x] Full suite: 119/119 passing.

### DIFF
--- a/packages/python/pyproject.toml
+++ b/packages/python/pyproject.toml
@@ -26,6 +26,7 @@ dependencies = [
     "numpy>=1.20",
     "trimesh>=3.0",
     "shapely>=1.8",
+    "defusedxml>=0.7.1",
 ]
 
 [project.optional-dependencies]
--- a/packages/python/src/openskp/_core.py
+++ b/packages/python/src/openskp/_core.py
@@ -18,7 +18,14 @@ import struct
 import time
 import zipfile
 from typing import Any, Dict
-import xml.etree.ElementTree as ET
+# material.xml/style.xml come from inside the .skp's ZIP container, which is
+# fully attacker-controlled input - stdlib ElementTree expands internal
+# general entities with no limit, making a "billion laughs" DoS payload
+# trivial to embed. defusedxml.ElementTree is a drop-in replacement (same
+# ParseError class, same fromstring() signature) that rejects that class of
+# payload outright instead.
+import defusedxml.ElementTree as ET
+from defusedxml.common import DefusedXmlException
 
 import numpy as np
 import trimesh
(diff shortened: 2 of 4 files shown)


iamahsanmehmood (commented):
Independently re-verified: full suite 119/119 passing (including a synthetic-.skp regression test embedding both a malicious material.xml and style.xml, confirming they're silently skipped rather than crashing/hanging the parse), `ruff check src/ tests/` clean, all 13 CI jobs green. Confirmed defusedxml blocks entity declarations outright (verified directly - even a single non-recursive declaration triggers `EntitiesForbidden` before any expansion), and that the widened except clauses correctly catch it alongside genuine `ET.ParseError`s. Merging.

PR #110, opened 2026-08-11
Add exact uv_transform regression count to legacy test (TypeScript)

iamahsanmehmood:
## Summary

Matches what C++'s `parser_test.cpp` already does (`with_uv_transform == 32` against `capilla_quiroz_v17.skp`). Only C++ would have caught a regression of the exact historical bug it already hit once: `CFace`'s attribute container (which carries any positioned/photo-fitted `CFaceTextureCoords` record) was read and then silently discarded, so `uvTransform` never got populated for legacy files at all - every legacy face fell back to the default face-plane projection even when the SketchUp author had explicitly positioned a texture.

## Changes

- Added a count of faces with `uvTransform` set (and `uvProjected`, for the same reason) across all definitions + root in `legacy.test.ts`'s existing real-fixture test.

## Test plan

- [x] Verified the counts directly against a fresh parse of the fixture before writing the assertion (32 / 0, matching Python's and C++'s independently-verified counts on this exact file - 0 have `uvProjected` since this file has no PROJECTED/terrain-drape textures).
- [x] Full suite: 65/65 passing.
- [x] `tsc --noEmit` clean.

### DIFF
--- a/packages/typescript/tests/legacy.test.ts
+++ b/packages/typescript/tests/legacy.test.ts
@@ -136,6 +136,30 @@ describe('Legacy MFC reader (classic pre-2021 .skp)', () => {
         expect(face.hidden).toBe(false);
       }
     }
+
+    // Regression: CFace's attribute container (which carries any
+    // positioned/photo-fitted CFaceTextureCoords record) was read and then
+    // silently discarded, so uvTransform never got populated for legacy
+    // files at all - every legacy face fell back to the default
+    // face-plane projection even when the SketchUp author had explicitly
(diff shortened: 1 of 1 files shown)


iamahsanmehmood (commented):
Independently re-verified: full suite 65/65 passing, `tsc --noEmit` clean, all CI jobs (Node 18/20/22) green. Confirmed the 32/0 counts against a fresh parse of the fixture before this was written, matching Python's/C++'s independently-verified counts. Merging.

PR #113, opened 2026-08-11
Add BuildScene/MeshIndex real-fixture coverage to .NET IntegrationTests.cs

iamahsanmehmood:
## Summary

TS/C++ both already exercise `BuildScene()`/`MeshIndex` against the real modern (VFF/2021+) fixtures (`Untitled.skp`, `SU_File.skp`) in their integration tests - .NET's real-fixture coverage stopped at `Parse()`, leaving the opt-in scene-baking path (`SceneBuilder`, mesh grouping, GLB-ready output) completely untested against a real file.

## Changes

Added the same assertions TS's `integration.test.ts` already makes, verified directly against a fresh `BuildScene()` call on both fixtures before writing them:
- `Untitled.skp`: `SceneHierarchy.Name == "ROOT"`, `SceneHierarchy.DefinitionName == "ROOT_MODEL"`, at least one child, `MeshIndex.Count == 43` (matches TS's/Python's/C++'s independently-verified count on this exact file).
- `SU_File.skp`: same `SceneHierarchy` checks, `MeshIndex` has exactly 1 entry whose `DefinitionName` is `"ROOT_MODEL"` (only ROOT holds geometry in this fixture).

## Test plan

- [x] Full suite: 46/46 passing.

### DIFF
--- a/packages/dotnet/OpenSkp.Tests/IntegrationTests.cs
+++ b/packages/dotnet/OpenSkp.Tests/IntegrationTests.cs
@@ -112,6 +112,21 @@ namespace OpenSkp.Tests
             Assert.Equal("[Construction Documentation Style]", model.Styles[0].Name);
             Assert.Equal((255, 255, 255), model.Styles[0].FrontColor);
             Assert.Equal((208, 209, 189), model.Styles[0].BackColor);
+
+            // BuildScene/MeshIndex - a separate, opt-in step from Parse(), so
+            // it never costs a plain Parse() call anything. TS/Dart/C++ all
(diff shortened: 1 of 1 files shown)


iamahsanmehmood (commented):
Independently re-verified: full suite 46/46 passing, clean build with 0 warnings, all CI jobs (macOS/ubuntu/windows) green. Confirmed the SceneHierarchy/MeshIndex values against a fresh BuildScene() call before this was written (43 meshes for Untitled.skp, 1 for SU_File.skp), matching TS's/Python's/C++'s independently-verified counts. Merging.

PR #115, opened 2026-08-11
Port double-sided back-face material handling to build_scene() (Python)

iamahsanmehmood:
## Summary

C++'s `scene.cpp` was the only port that correctly rendered a face whose front and back are painted with genuinely different materials: it splits the face into two single-sided triangle sets (front normal-wound, back reverse-wound), each keeping its own material - so a wall painted white on one side and green on the other actually renders both colors. Python (like TS/Dart/.NET) only ever resolved the FRONT material for the whole face, baked one single-sided mesh with normal winding, and never looked at `back_material_id`/`uv_transform_back` at all - so the back side either showed the wrong (front) color or, with backface culling on (the default - no port ever set glTF's `doubleSided` flag), was invisible outright.

This is a real, verifiable defect: the legacy fixture (`capilla_quiroz_v17.skp`) has **30 faces with a front/back color that genuinely differs**.

## Changes

- `scene.py`: `build_scene()`'s face-processing loop now resolves both `front_mat` (`material_id`) and `back_mat` (`back_material_id`) per face.
  - When their resolved colors match (including the common case of no back material at all), the face is emitted once as before, but its glTF material is now marked `doubleSided=True` - a genuine improvement even in the "colors match" case, since previously NO material was ever double-sided, meaning a face could still vanish when viewed from behind for no reason.
  - When the colors differ, the face is split into two single-sided primitives: front (normal winding, front material/`uv_transform`/texture), back (reverse winding - swap triangle indices 1 and 2 - back material/`uv_transform_back`/texture, negated accumulated normal).
  - Mesh/material grouping keys extended from `color` to `(color, double_sided)` to keep these two variants separate. Faithfully mirrors C++'s `GroupKey{color, double_sided}` approach.

## Test plan

- [x] Verified against the real legacy fixture: `glb_primitives`/`mesh_index`/`gltf_materials` counts (21/21/13) now match C++'s own independently-verified reference for this exact file exactly (`parser_test.cpp`), confirming the port is correct rather than just "different."
- [x] New `TestBuildScene.test_back_face_materials_render_correctly` asserts two specific real faces' front (blue, `(2,0,237)`) and back (light blue, `(204,235,244)`) colors both appear as distinct `gltf_materials`, and that exactly 4 materials ended up marked `doubleSided` on this fixture.
- [x] Existing `TestBuildScene` assertions updated from the old (incorrect, front-only) counts (13/13/9) to the new correct ones; the class docstring now explains why these no longer match TypeScript's still-unported counts (a Python-only PR - TS/Dart/.NET need the same port as follow-ups).
- [x] Fixed a now-invalid assertion in `TestGlbExport` that compared the separate trimesh-based `export()` pipeline's primitive count against `build_scene()`'s - that equality was only ever a coincidence, not a documented invariant (removed with an explanatory comment; the test's real substance, real UV per exported mesh, is unaffected).
- [x] Full suite: 122/122 passing.

### DIFF
--- a/packages/python/src/openskp/scene.py
+++ b/packages/python/src/openskp/scene.py
@@ -65,9 +65,11 @@ class MeshMetadata:
 
 @dataclass
 class GlbPrimitive:
-    """One triangulated, world-space mesh: all faces sharing a single
-    resolved color from one flattened scene-graph position. Ready to hand
-    straight to a GLB/glTF exporter or any other renderer.
+    """One triangulated, world-space mesh: all faces (or, for a face whose
+    front/back colors genuinely differ, all *one side* of those faces)
+    sharing a single resolved color from one flattened scene-graph
+    position. Ready to hand straight to a GLB/glTF exporter or any other
+    renderer.
 
     Attributes:
         positions: Flat [x, y, z, x, y, z, ...] vertex positions, in
@@ -213,7 +215,7 @@ def build_scene(parsed: Dict[str, Any]) -> Scene:
     mesh_index: Dict[str, MeshMetadata] = {}
     glb_primitives: List[GlbPrimitive] = []
 
-    color_to_material_index: Dict[Tuple[int, int, int], int] = {}
+    color_to_material_index: Dict[Tuple[Tuple[int, int, int], bool], int] = {}
     gltf_materials: List[Dict[str, Any]] = []
 
     # Definitions currently being instantiated on the active recursion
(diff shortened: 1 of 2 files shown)


iamahsanmehmood (commented):
Independently re-verified: full suite 122/122 passing, `ruff check src/ tests/` clean, all 13 CI jobs green. Confirmed the exact new glb_primitives/mesh_index/gltf_materials counts (21/21/13) precisely match C++'s own independently-verified reference for this fixture, not just a plausible-looking different number. Confirmed two specific real faces' front (blue) and back (light blue) colors both appear as distinct materials, and exactly 4 materials ended up doubleSided. The now-invalid trimesh-vs-build_scene() primitive-count equality was correctly identified as a coincidence (not a documented invariant) and removed with a clear explanation rather than papered over. Merging.

PR #116, opened 2026-08-11
Port double-sided back-face material handling to buildScene() (TypeScript)

iamahsanmehmood:
## Summary

Companion to #115 (Python). C++'s `scene.cpp` was the only port that correctly rendered a face whose front and back are painted with genuinely different materials: it splits the face into two single-sided triangle sets (front normal-wound, back reverse-wound), each keeping its own material - so a wall painted white on one side and green on the other actually renders both colors. TypeScript (like Python/Dart/.NET) only ever resolved the FRONT material for the whole face, baked one single-sided mesh with normal winding, and never looked at `backMaterialId`/`uvTransformBack` at all - so the back side either showed the wrong (front) color or, with backface culling on (the default - no port ever set glTF's `doubleSided` flag), was invisible outright.

This is a real, verifiable defect: the legacy fixture (`capilla_quiroz_v17.skp`) has **30 faces with a front/back color that genuinely differs**. Unlike Python, TypeScript's `buildScene()` output IS what its own `toGLB()`/`exportGlb()` actually write to disk - no disconnected-pipeline caveat here, this fix changes real exported `.glb` output.

## Changes

- `model.ts`: `buildScene()`'s face-processing loop now resolves both `frontMat` (`materialId`) and `backMat` (`backMaterialId`) per face.
  - When their resolved colors match (including the common case of no back material at all), the face is emitted once as before, but its glTF material is now marked `doubleSided=true`.
  - When the colors differ, the face is split into two single-sided primitives: front (normal winding, front material/`uvTransform`/texture), back (reverse winding - swap triangle indices 1 and 2 - back material/`uvTransformBack`/texture, negated accumulated normal).
  - Mesh/material grouping keys extended from `color` to `(color, doubleSided)` to keep these two variants separate. Faithfully mirrors C++'s `GroupKey{color, doubleSided}` approach and Python's just-merged `addSide()` structure.

## Test plan

- [x] Verified against the real legacy fixture: `glbPrimitives`/`meshIndex`/`gltfMaterials` counts (21/21/13) now match C++'s and Python's independently-verified reference for this exact file exactly.
- [x] New `legacy.test.ts` assertion (item 14 regression) verifies two specific real faces' front (blue, `(2,0,237)`) and back (light blue, `(204,235,244)`) colors both appear as distinct `gltfMaterials` entries, that exactly 4 materials ended up `doubleSided`, and the full `glbPrimitives`/`meshIndex`/`gltfMaterials` counts.
- [x] No existing test assertions needed updating - `legacy.test.ts`'s `build_scene()` coverage never hardcoded exact primitive/material counts (unlike Python's, which did and needed updating in #115), and `glb.test.ts`'s real-fixture GLB comparison derives its expected count from `scene.glbPrimitives.length` dynamically rather than a hardcoded number.
- [x] Full suite: 66/66 passing.
- [x] `tsc --noEmit` clean.

### DIFF
--- a/packages/typescript/src/model.ts
+++ b/packages/typescript/src/model.ts
@@ -166,9 +166,10 @@ export interface MeshMetadata {
   path: string;
 }
 
-/** One triangulated, world-space mesh: all faces sharing a single resolved
- * color from one flattened scene-graph position. Ready to hand straight to
- * a GLB/glTF exporter or any other renderer. */
+/** One triangulated, world-space mesh: all faces (or, for a face whose
+ * front/back colors genuinely differ, all *one side* of those faces)
+ * sharing a single resolved color from one flattened scene-graph position.
+ * Ready to hand straight to a GLB/glTF exporter or any other renderer. */
 export interface GlbPrimitive {
   /** Flat [x, y, z, x, y, z, ...] vertex positions, in metres, Y-up. */
   positions: Float32Array;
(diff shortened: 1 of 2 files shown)


iamahsanmehmood:
Independently re-verified before merge: `npx vitest run tests/legacy.test.ts` passes, including the new back-face regression test asserting the two real resolved colors (2,0,237 front / 204,235,244 back) on `puerta` faces 133/152 and the exact 21/21/13/4 primitive/meshIndex/material/doubleSided counts. These match C++'s own pre-existing reference for this fixture (`EXPECT_EQ(scene.glb_primitives.size(), 21)` etc. in parser_test.cpp) and the already-merged Python port (#115). CI green across all three test matrix jobs. Merging.

PR #117, opened 2026-08-11
Port double-sided back-face material handling to buildScene() (Dart)

iamahsanmehmood:
## Summary

Dart's `scene.dart` resolved only the front material per face, discarding `backMaterialId` entirely - a face painted with genuinely different front and back materials rendered with the front color on both sides, and no glTF material was ever marked `doubleSided` even in the shared-color case.

Ports C++'s `scene.cpp` reference behavior (the only language that already handled this correctly) - see the already-merged Python (#115) and TypeScript (#116) ports of the same fix for prior art:

- Resolve front/back materials independently, falling back to the same inherited/layer color for whichever side lacks a material.
- Same resolved color on both sides → emit one triangle set, mark the glTF material `doubleSided: true`.
- Different colors → split into two single-sided triangle sets: front (normal winding, front material/uvTransform) and back (reversed winding - swap triangle indices 1 and 2 - negated per-vertex normals, back material/uvTransformBack).
- `_FaceGroup`/`colorToMaterialIndex` keyed on `(color, doubleSided)` so same-color and split variants stay separate; geom-name color suffix gains a `ds`/`ss` marker when more than one group exists.

## Verification

Against the same real fixture used by the Python/TS ports: `capilla_quiroz_v17.skp` has 30 faces with genuinely different front/back render colors (e.g. `puerta` faces 133/152: front material 29 = blue `(2,0,237)`, back material 27 = light blue `(204,235,244)`).

- `glbPrimitives`/`meshIndex` count: 13 → 21
- `gltfMaterials` count: 9 → 13
- `doubleSided` materials: 4

These match C++'s independently-verified reference counts for this exact file (`parser_test.cpp`'s `EXPECT_EQ(scene.glb_primitives.size(), 21)` / `EXPECT_EQ(scene.gltf_materials.size(), 13)`), and the already-merged Python/TS counts exactly.

Added a dedicated regression test asserting both real resolved colors are present and the doubleSided count is 4.

- `dart analyze`: clean, no issues.
- `dart test`: all passing, including the updated `scene_test.dart` counts and the new back-face regression test.

## Test plan

- [x] `dart analyze` clean
- [x] `dart test` all green (updated ground-truth counts + new regression test)
- [x] Cross-checked exact counts against C++'s pre-existing reference and Python/TS's already-merged ports

### DIFF
--- a/packages/dart/lib/src/scene.dart
+++ b/packages/dart/lib/src/scene.dart
@@ -106,12 +106,13 @@ class Scene {
 
 class _FaceGroup {
   final (int, int, int) color;
+  final bool doubleSided;
   final List<(double, double, double)> localVerts = [];
   final List<(double, double)> localUvs = [];
   final List<List<double>> normalsAccum = [];
   final List<List<int>> localFaces = [];
   final Map<(int, double, double), int> localVMap = {};
-  _FaceGroup(this.color);
+  _FaceGroup(this.color, this.doubleSided);
 }
 
 const double _inchesToMm = 25.4;
@@ -195,7 +196,7 @@ class SceneBuilder {
     final meshIndex = <String, MeshMetadata>{};
     final glbPrimitives = <GlbPrimitive>[];
 
-    final colorToMaterialIndex = <(int, int, int), int>{};
+    final colorToMaterialIndex = <((int, int, int), bool), int>{};
     final gltfMaterials = <Map<String, dynamic>>[];
 
     // Definitions currently being instantiated on the active recursion path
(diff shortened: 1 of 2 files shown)


iamahsanmehmood:
Independently re-verified before merge: `dart analyze` clean, `dart test` all green including the updated ground-truth counts (13→21 primitives/meshIndex, 9→13 materials) and the new back-face regression test (both real resolved colors present, doubleSided count = 4). These exactly match C++'s pre-existing reference for this fixture and the already-merged Python (#115) and TypeScript (#116) ports. CI green on macOS/Ubuntu/Windows. Merging.

PR #118, opened 2026-08-11
Port double-sided back-face material handling to BuildScene() (.NET)

iamahsanmehmood:
## Summary

.NET's `Scene.cs` resolved only the front material per face, discarding `BackMaterialId` entirely - a face painted with genuinely different front and back materials rendered with the front color on both sides, and no glTF material was ever marked `doubleSided` even in the shared-color case.

Ports C++'s `scene.cpp` reference behavior (the only language that already handled this correctly), following the same pattern already shipped for Python (#115), TypeScript (#116), and Dart (#117):

- Resolve front/back materials independently via a new `ResolveMaterial()` helper, falling back to the same inherited/layer color for whichever side lacks a material.
- Same resolved color on both sides → emit one triangle set via a new `AddSide()` local function, mark the glTF material `doubleSided: true`.
- Different colors → split into two single-sided triangle sets: front (normal winding, front material/UvTransform) and back (reversed winding - swap triangle indices 1 and 2 - negated per-vertex normals, back material/UvTransformBack).
- `FaceGroup`/`colorToMaterialIndex` keyed on `(Color, DoubleSided)` so same-color and split variants stay separate; geom-name color suffix gains a `ds`/`ss` marker when more than one group exists.
- `GltfMaterials` entries switch from anonymous types to `Dictionary<string, object>` so a `doubleSided` key can be added conditionally - `MiniJson.Write` already special-cases `IDictionary<string, object>` ahead of its reflection fallback, so GLB serialization is unaffected.

## Verification

Against the same real fixture used by the Python/TS/Dart ports: `capilla_quiroz_v17.skp` has 30 faces with genuinely different front/back render colors (e.g. `puerta` faces 133/152: front material 29 = blue `(2,0,237)`, back material 27 = light blue `(204,235,244)`).

- `GlbPrimitives`/`MeshIndex` count: 13 → 21
- `GltfMaterials` count: 9 → 13
- `doubleSided` materials: 4

These match C++'s independently-verified reference counts for this exact file, and the already-merged Python/TS/Dart counts exactly.

Added a dedicated regression test asserting both real resolved colors are present and the doubleSided count is 4.

- `dotnet build`: 0 warnings, 0 errors.
- `dotnet test`: 47/47 passing, including the updated `SceneTests` counts, the new back-face regression test, and the untouched `Untitled.skp` `IntegrationTests` fixture (its MeshIndex count of 43 is unaffected - confirmed it has no differing front/back faces).

## Test plan

- [x] `dotnet build` clean
- [x] `dotnet test` all green (47/47, updated ground-truth counts + new regression test)
- [x] Cross-checked exact counts against C++'s pre-existing reference and Python/TS/Dart's already-merged ports

### DIFF
--- a/packages/dotnet/OpenSkp.Tests/SceneTests.cs
+++ b/packages/dotnet/OpenSkp.Tests/SceneTests.cs
@@ -11,11 +11,15 @@ namespace OpenSkp.Tests
     /// scene-hierarchy + triangulation + GLB mesh capability, ported from
     /// the TypeScript reference implementation.
     ///
-    /// Cross-validated directly against Python's and TypeScript's
-    /// SkpFile.build_scene()/buildScene() on this exact fixture: mesh
-    /// count, mesh_index count, gltf_materials count, root instance count,
-    /// and the first three meshes' exact vertex/triangle counts and
-    /// material indices all match precisely.
+    /// Root instance count is cross-validated directly against Python's
+    /// and TypeScript's SkpFile.build_scene()/buildScene() on this exact
+    /// fixture. Mesh/GltfMaterials counts (21/21/13) instead match C++'s
+    /// independently-verified reference for this file - the correct
+    /// counts once faces with genuinely different front/back materials
+    /// are split into two single-sided primitives each, rather than the
+    /// pre-fix single-sided-only count (13/13/9). This fixture has 30
+    /// such faces (confirmed by direct inspection), so the split isn't a
+    /// rare edge case here.
     /// </summary>
     public class SceneTests
     {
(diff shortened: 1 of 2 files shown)


iamahsanmehmood:
Independently re-verified before merge: `dotnet build` clean (0 warnings, 0 errors), `dotnet test` 47/47 passing including the updated SceneTests counts (13→21 primitives/meshIndex, 9→13 materials), the new back-face regression test (both real resolved colors present, doubleSided count = 4), and the untouched Untitled.skp IntegrationTests fixture (its MeshIndex count of 43 confirmed unaffected). These exactly match C++'s pre-existing reference for this fixture and the already-merged Python (#115)/TypeScript (#116)/Dart (#117) ports. CI green on macOS/Ubuntu/Windows plus the NuGet build-and-publish job. Merging - this completes item 14 across all 5 languages.

PR #119, opened 2026-08-11
Wire Python's GLB export to scene.build_scene() instead of a separate pipeline

iamahsanmehmood:
## Summary

Resolves audit item 15. `openskp.export.glb.export()` ran its own ~270-line reimplementation of `scene.py`'s instantiation/triangulation/material-resolution logic directly against a live `trimesh.Scene`, instead of using `scene.build_scene()` (the same baking step `SkpFile.build_scene()` exposes, and the one every other language's real GLB export is built on).

That duplication had already drifted out of sync in two concrete, user-visible ways:

- **No recursion/cycle guard.** `scene.py` was hardened against a component definition that directly or transitively instances itself (item 24 - raises `SkpParseError` instead of recursing until the stack overflows), but `export.glb.export()`'s separate `instantiate()` walk never got the same fix - a malicious or corrupt `.skp` file with a self-referencing component would crash the actual export path.
- **No back-face material handling (item 14).** `export.glb.export()` only ever resolved a face's front `material_id`, never `back_material_id`/`uv_transform_back` - so real exported `.glb` files still rendered the front color on both sides of a face, even after `scene.build_scene()` was fixed to resolve front/back independently across all 5 languages.

## Change

`export/glb.py` now calls `scene.build_scene()` to bake the scene, converts each `GlbPrimitive` to a `trimesh.Trimesh` (position/normal/uv arrays used directly - UV-based vertex splitting is already done, `process=False` keeps it that way) with a `PBRMaterial` built from the primitive's resolved `gltf_materials` entry (`baseColorFactor`/`metallicFactor`/`roughnessFactor`/`doubleSided` all carry straight through), and hands the assembled `trimesh.Scene` to trimesh purely for GLB binary serialization.

The JSON metadata sidecar and PNG thumbnail extraction are preserved, rebuilt from the baked `Scene`'s `mesh_index`/`scene_hierarchy` instead of bespoke locally-built dicts. Two fields (`definition_id`, `matrix_3x4`) are dropped from the per-instance metadata JSON since they aren't part of `InstanceNode`'s cross-language-parity shape and would have required re-deriving via a second walk, defeating the point of removing the duplicate implementation.

`scene.build_scene()`'s `GlbPrimitive` positions are in metres (matching every other language's `Scene`/`GlbPrimitive`) - `export()`'s own public contract has always been millimetres (`units="mm"` is the only supported value), so positions are scaled by 1000 in the trimesh conversion step to keep that contract unchanged for existing callers, rather than silently changing the scale of files they already produce.

`_core.py`'s old `build_scene()` (270 lines, the only thing importing `trimesh` in that module) is deleted outright - nothing else referenced it. `import json` also dropped from `_core.py` as it was only used there.

## Verification

- Independently confirmed the millimetre scaling is correct: computed `capilla_quiroz_v17.skp`'s bounding box directly from `scene.build_scene()` (metres) and compared against the exported `.glb`'s bounds as loaded by trimesh - matched to within float32 precision (ratio 0.99999997).
- Independently confirmed `doubleSided` now reaches real output: parsed the exported `.glb`'s raw JSON chunk by hand (not just via trimesh's Python objects) - 13 materials, 4 marked `doubleSided`, matching item 14's Python/TS/Dart/.NET ground truth exactly.
- Added `test_export_rejects_self_referencing_definition`, confirming the recursion guard now reaches `export.glb.export()` itself (it didn't before this change).
- Added `test_export_marks_back_face_materials_double_sided`, asserting the same 13/4 material counts by parsing the real exported `.glb` bytes.
- Updated `test_metadata_json_is_written_and_json_safe`'s `total_meshes` from 13 to 21 (now matches `TestBuildScene`'s ground truth, since both paths bake through the same function) and its per-instance key assertions for the two dropped fields.

Full suite: 124/124 passing. `ruff check src/ tests/`: clean.

## Test plan

- [x] `pytest tests/` 124/124 passing
- [x] `ruff check src/ tests/` clean
- [x] Independently verified millimetre scale correctness against the raw baked Scene
- [x] Independently verified doubleSided flag reaches the raw GLB JSON bytes
- [x] Independently verified recursion guard now reaches export.glb.export()


iamahsanmehmood:
Independently re-verified before merge:
- `pytest tests/` 124/124 passing, `ruff check src/ tests/` clean.
- Computed capilla_quiroz_v17.skp's bbox directly from scene.build_scene() (metres) and compared against the exported .glb's bounds loaded via trimesh - matched to float32 precision (ratio 0.99999997), confirming the mm-scaling contract is preserved.
- Hand-parsed the exported .glb's raw JSON chunk (not just trimesh's Python objects): 13 materials, 4 doubleSided - matches item 14's cross-language ground truth exactly, confirming the back-face fix now reaches real Python .glb output.
- Confirmed the recursion guard now reaches export.glb.export() itself via the new test_export_rejects_self_referencing_definition (it only covered scene.build_scene() before this change).

CI green across the full 12-job OS/Python matrix plus lint. Merging - this closes item 15 (wired together, per your earlier direction) and also means item 14's fix now has real effect on Python's actual .glb output, resolving the caveat noted when #115 merged.

PR #155, opened 2026-08-15
fix(typescript): read SketchUp 2020 (v20) classic .skp files

thomazyujibaba:
## Summary

Real SketchUp 2020 `.skp` files fail the legacy (pre-2021 MFC) walk with `implausible definition count`. v20 writes records the v17 layout does not have, so the reader ends up a few bytes short and reads garbage where a count is expected.

The only legacy fixture in the repo (`capilla_quiroz_v17.skp`) is v17 with a single layer, so none of these paths were exercised.

## Type
- [x] Bug fix

## Changes

Four divergences, each found by reading the bytes around the failure point:

**1. Empty-string filler ahead of several counts.** v20 writes an empty UTF-16 string record followed by zero padding before the definition count, the per-definition entity count, the relationship count, and the root entity count. `retryCountAfterV20Filler` only runs once a count already came back out of range, locates the marker, skips the padding and re-reads. Files that parsed before never reach it.

**2. A zero count is the same filler, seen from the other side.** The reader can also land on the filler's *leading* zero bytes, reading `count = 0` and returning an empty definition. This one is silent, and worse than the crash: on a second v20 file (a 47-definition architectural model I could not include here), 22 of the 47 definitions read a zero count — and they were exactly the ones the placed instances reference. The file parsed "successfully" into 178 correctly-positioned but completely empty groups: **1 baked primitive instead of 837**, with the definitions that actually carried the geometry left unreferenced. Since `retryCountAfterV20Filler` refuses to move the cursor unless the empty-string marker is really there, a genuinely empty definition still reads zero and is left alone.

**3. Null object-ref in the layer list.** `layerCount` includes a null ref that carries no layer record. It was pushed into `layers` as a `null` and blew up downstream on `v.rgba`.

**4. Two extra bytes before the per-definition GUID.** This shifted the read and left the cursor mid-record. The read now anchors on the string marker that must follow the 16 GUID bytes, rather than trusting a fixed prefix width — which works for both layouts.

## Testing

New fixture `gondola_v20.skp` (SketchUp 20.1.235, 755 KB, a retail gondola display) and `tests/legacy-v20.test.ts`.

The tests assert the instances are **actually placed** (23 root instances, 201 baked primitives, bounding box plausible in metres) rather than only that the parse stops throwing. A subtly misaligned walk yields an almost-empty scene instead of an error, which is worse than failing outright.

- 90/90 tests pass (`npm test`), `tsc --noEmit` clean
- v17 and VFF fixtures produce identical results to before
- Confirmed the new tests fail on `main` without the fix (4 of 5 cases)

**One gap worth flagging:** `gondola_v20.skp` exercises the zero-count path (its walk takes the retry 20 times) but does not *depend* on it — it parses identically with or without change #2. The file that does depend on it is the one I could not include, so that test asserts the invariant instead: every referenced definition must resolve to one carrying geometry, directly or through nested children. Counting definitions or instances alone does not catch the failure, since both come out correct while the scene is empty. If you have a v20 file in your corpus that reproduces the empty-scene case, it would make a stronger fixture than the invariant check.

## Checklist
- [x] Tests pass locally
- [x] No unrelated changes included

## Notes for other platforms

Per CONTRIBUTING's cross-platform parity note, the expected behavior for the Python/Dart/.NET/C++ ports: all four divergences live in the legacy walk, at the same call sites. The filler is `<ff fe ff><u8 0>` plus zero padding, and triggers on both an out-of-range count and a zero one; the layer null-ref should be skipped rather than stored; the GUID read should anchor on the trailing string marker. I only have TypeScript set up here, so I have not ported them.


iamahsanmehmood (approved):
Reviewed the diff and traced through the byte-level logic in `legacy.ts`. Approving.

A few notes:

- The layer null-ref fix and the GUID-anchor fix both look correct and match how the rest of the legacy reader already anchors on markers instead of trusting fixed offsets.
- `retryCountAfterV20Filler`'s "skip zero padding, first non-zero byte is the count's low byte" heuristic has a narrow edge case: if the real count is an exact multiple of 256 (its own low byte is 0x00), the scan can't distinguish that from filler padding and will misalign the read. In practice this mostly degrades to hitting the existing `> 5_000_000` / `> 100_000` / `> 1_000_000` implausibility checks and throwing cleanly rather than corrupting silently, so it's not a regression against the current behavior - just worth a follow-up if a v20 file ever trips it.
- CI wasn't running on this PR (fork PRs need workflow approval on first contribution) - approved the run manually and it's green across Node 18/20/22 plus lint.
- Confirmed this is isolated to `packages/typescript/src/legacy.ts` + the new fixture/test, no unrelated changes.

Nice regression tests, especially checking that referenced definitions actually carry geometry rather than just counting definitions/instances - that's exactly the invariant that would have caught the empty-scene failure mode.

PR #174, opened 2026-08-18
fix: geometry loses its material in two cases (ZIP entry names, instance-painted faces)

vvdimoff:
Two independent bugs, both of which end with geometry losing its material — and both found while building a browser viewer for `.skp` on top of openskp, against real customer furniture models.

---

# 1. Materials parsed with `id === null` (ZIP entry names)

SketchUp stores entry names inside the `.skp` archive as UTF-8 but leaves the ZIP language encoding flag (general purpose bit 11) clear. A spec-compliant reader therefore decodes them as CP437/Latin-1, so a material folder named `ДСП Egger` arrives as `ÐÐ¡Ð Egger`.

That folder name is the key of `materialsByFolder`, and `model.ts` joins TLV material ids onto materials by name:

```ts
for (const [mId, mName] of materialIdToName.entries()) {
  const mat = materialsMap.get(mName) || materialsByFolder.get(mName);
  if (!mat) continue;          // <- mangled key never matches
  if (mat.id === null) mat.id = mId;
  materialsById.set(mId, mat);
}
```

The mangled key never matches, so the material keeps `id: null` and its TLV id is missing from `materialsById`. Every face referencing that id is then treated as material-less and falls back to the layer colour.

Worth noting the name in the TLV record is *also* truncated (`$69498 ДСП Egger F 204 ST9 Мармур Каррара`, cut at 57 bytes, against the full `… білий 2800х2070х18 мм` in `material.xml`), so `materialsMap` cannot match it either — the folder key is the only join that can work here, which is exactly the one the encoding breaks.

**Fix:** recover the original bytes and decode them again as UTF-8. The Latin-1 mapping is lossless, so the round trip is exact; names that are not valid UTF-8 are kept as they are, and pure ASCII names pass through untouched.

| | before | after |
|---|---|---|
| entries in `materialsById` | 12 | 13 |
| face `materialId 770085` resolves | no | yes |
| countertop | tag colour, no texture | its own material + texture |
| countertop UV span | 59.06 (raw inches) | 2.46 repeats over a 23.4″ part with a 24″ tile |

---

# 2. A face inherits only the colour of the instance material, not the material

A face with no material of its own is painted by the instance around it (SketchUp's "paint the component"). `instantiate()` resolves that inheritance for the **colour** only — it passes `inheritedMaterialColor` down the recursion — so `addSide` still receives `undefined` for the material and `computeFaceUv` falls back to `tileW = tileH = 1`:

```ts
const tex = mat?.texture;
const tileW = tex && tex.width > 1e-9 ? tex.width : 1;
const tileH = tex && tex.height > 1e-9 ? tex.height : 1;
```

The UVs of every such face therefore come out in raw inches. This is not an exotic case: in one of the affected files **not a single face** references a textured material — every panel is painted at the group level, which is how furniture models normally come out of a cabinet plugin.

Measured on that file:

| material | tile | part | UV span before | UV span after |
|---|---|---|---|---|
| `$154335 ДСП Swiss Krono 4413` | 46.89″ | 21.7″ | 40.45 (1.87 repeats per inch — a pattern every 13 mm) | 0.86 |
| `$171665 МДФ Arpa 4625` | 75.59″ | 23.5″ | 10.00 | 0.13 |

**Fix:** thread the whole material through the recursion instead of only its colour, and let a face with no material of its own inherit it. The colour keeps working exactly as before, since it is read from the same material.

---

## Tests

* `tests/zip-entry-name-encoding.test.ts` — builds synthetic archives the way `zip-entry-size-cap.test.ts` does: a non-ASCII folder name is restored, an ASCII name is untouched, a name that is not valid UTF-8 is left alone.
* `tests/instance-material-uv.test.ts` — builds a synthetic parsed model (one unpainted square panel inside a painted instance) and asserts the UV span equals panel/tile in both axes. On the previous behaviour it reports 24 repeats instead of 1.

Full suite passes (98 tests), plus `npm run lint` and `npm run typecheck`.

## Notes

The affected models are customer files, so I cannot attach one publicly — happy to share privately if you want to reproduce against the originals. Any `.skp` with non-ASCII material names hits the first bug; any file where groups are painted rather than faces hits the second.

Thanks for openskp. It does a remarkable job on files nothing else opens without SketchUp, and the code was a pleasure to read while tracking these down.

### DIFF
--- a/packages/typescript/src/model.ts
+++ b/packages/typescript/src/model.ts
@@ -512,7 +512,7 @@ export function buildSceneFromParsed(parsed: ParsedRawData, options?: ParseOptio
     currentMatrix: number[],
     parentLayer: string = 'Layer0',
     pathName: string = 'ROOT',
-    inheritedMaterialColor?: { r: number; g: number; b: number }
+    inheritedMaterial?: Material
   ): InstanceNode[] {
     const d = defsDict.get(defId);
     if (!d) return [];
@@ -618,10 +618,16 @@ export function buildSceneFromParsed(parsed: ParsedRawData, options?: ParseOptio
       };
 
       for (const [_fId, fData] of builder.faces.entries()) {
-        const fallbackColor = inheritedMaterialColor ?? getLayerColor(parentLayer);
+        const fallbackColor = inheritedMaterial?.color ?? getLayerColor(parentLayer);
 
-        const frontMat = resolveMaterial((fData as any).materialId);
-        const backMat = resolveMaterial((fData as any).backMaterialId);
+        // A face with no material of its own is painted by the instance
+        // (SketchUp's "paint the component"). Inheriting the whole material -
+        // not just its colour - is what gives computeFaceUv the texture's tile
+        // size: without it tileW/tileH fall back to 1 and the UVs come out in
+        // raw inches, so a 1.9 m decor sheet tiles ~150 times across a 600 mm
+        // panel instead of covering a third of it.
+        const frontMat = resolveMaterial((fData as any).materialId) ?? inheritedMaterial;
+        const backMat = resolveMaterial((fData as any).backMaterialId) ?? inheritedMaterial;
         const frontColor = frontMat?.color ?? fallbackColor;
         const backColor = backMat?.color ?? fallbackColor;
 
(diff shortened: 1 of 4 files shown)


iamahsanmehmood (approved):
Verified locally: full TypeScript suite passes (98/98, including the 2 new test files), tsc --noEmit clean, and both new test files are eslint-clean under the project's own lint scope. The two fixes (ZIP entry names decoded as Latin-1 instead of UTF-8; instance-painted faces inheriting only the material's color instead of the full material, losing texture tile size for UV scaling) are correctly diagnosed and each has regression coverage, including edge cases (ASCII names untouched, invalid-UTF-8 names left alone, unpainted instances). All required CI is green.

iamahsanmehmood:
Thanks for the fix and the thorough regression tests — both bugs (ZIP entry name decoding, instance-inherited material) were well diagnosed and clearly explained. Merged.

PR #175, opened 2026-08-18
fix(dotnet): make BuildScene instance→mesh back-fill O(n) instead of O(n²)

MoCongxin:
## Summary

`SceneBuilder.Build` re-scans the **entire** `meshIndex` for **every** placed
instance with a `string.Contains` per mesh, making `BuildScene` O(instances ×
meshes). On a 183 MB production file (323,856 placed instances × 321,683
meshes) this single loop took **~73 minutes** of a ~75-minute conversion —
≈98% of the whole pipeline. The fix records each instance’s (Properties,
Name) in O(1) during instantiation and applies the updates once afterwards
by walking each mesh’s `Path` to its shallowest recorded ancestor
(O(meshes × path_depth)), preserving the existing final-state semantics.

## Type

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

## Changes

**`packages/dotnet/OpenSkp/Scene.cs`** (`SceneBuilder.Build`):

- Replace the per-instance full `meshIndex` scan (string `Contains` per mesh)
  with an O(1) `pathUpdates[fullPathName] = (properties, name)` registration
  during instantiation.
- Add a post-instantiation pass that walks each mesh’s `MeshMetadata.Path`
  and applies the shallowest (outermost) recorded ancestor update — the same
  final state the old code produced (the outermost instance wrote last).
- Incidental fix: the old substring test could over-match — a path like
  `ROOT / A_B` matched any GeomName merely *containing* `A_B` (e.g. inside
  `A_BC`), propagating an update to non-descendant meshes. The
  Path-prefix walk cannot.
- Public API is unchanged (`SceneBuilder.Build` is internal); output is
  byte-identical for existing models.

**`packages/dotnet/OpenSkp.Tests/SceneInstancePropagationTests.cs`** (new):

- `MeshBackfill_OutermostAncestorWins` — pins semantics: nested instances,
  outermost ancestor wins, root mesh untouched (synthetic `RawParsed`, no
  fixture needed).
- `MeshBackfill_LargeInstanceCount_IsLinear` — 100k placed instances complete
  in ~1 s with the fix; **fails on the old code** (~2.5 min, over the 30 s
  bound), guarding against regression to quadratic behaviour.

## Testing

- Full .NET test suite: **69/69 pass** (67 pre-existing + 2 new).
- `dotnet format --verify-no-changes` passes for both `OpenSkp` and
  `OpenSkp.Tests` projects (matches the CI `format` job).
- `dotnet build -warnaserror` (analyzers as errors) passes with 0 warnings.
- Downstream consumer benchmark (183 MB SKP):

  | | before | after |
  | --- | --- | --- |
  | BuildScene (build_scene stage) | 4374.6 s | 32.5 s (**135×**) |
  | total conversion | 4481.5 s | 103 s (**43×**) |
  | output (.bin) | — | content-identical (all tables/geometry hashes match) |

- The 100k-instance linearity test was validated to fail on the pre-fix code
  (2m38s) and pass on the fixed code (~1s).

## Checklist

- [x] Code follows the project’s style guidelines (`dotnet format` clean;
  analyzers pass with `-warnaserror`)
- [x] Tests pass locally (69/69, saw the new tests fail on the old code)
- [x] Documentation updated (if applicable) — N/A: internal-only change, no
  public API or behavior change; the only behavior difference is the
  elimination of a latent non-descendant over-match

---

## Note (follow-up, not in this PR)

The identical O(n²) scan exists in the TypeScript reference
(`packages/typescript/src/model.ts`, `buildSceneFromParsed`) and in Python
(`packages/python/src/openskp/scene.py`), but the current PR hasn't been processed yet.

### DIFF
--- a/packages/dotnet/OpenSkp.Tests/SceneInstancePropagationTests.cs
+++ b/packages/dotnet/OpenSkp.Tests/SceneInstancePropagationTests.cs
@@ -1,0 +1,150 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using Xunit;
+using OpenSkp;
+
+namespace OpenSkp.Tests
+{
+    /// <summary>
+    /// Regression tests for the deferred instance -> mesh (Properties, Name)
+    /// back-fill in SceneBuilder.Build.
+    ///
+    /// The old implementation re-scanned the entire meshIndex with a string
+    /// Contains per placed instance, i.e. O(instances x meshes). On real
+    /// production files with a few hundred thousand instances and meshes
+    /// that single loop took tens of minutes (measured: 323,856 instances x
+    /// 321,683 meshes -> ~73 minutes in BuildScene, ~98% of the total
+    /// conversion time). The replacement records one dictionary entry per
+    /// instance during instantiation and applies updates once afterwards by
+    /// walking each mesh's Path to its shallowest recorded ancestor.
+    ///
+    /// These tests build synthetic RawParsed models directly (internals are
+    /// visible to the test assembly) so the semantics and the linear-time
+    /// behaviour are pinned without needing a large .skp fixture.
+    /// </summary>
+    public class SceneInstancePropagationTests
+    {
+        private static GeometryBuilder TriangleBuilder(long vidBase)
+        {
+            var b = new GeometryBuilder();
+            b.Vertices[vidBase + 0] = (0.0, 0.0, 0.0);
+            b.Vertices[vidBase + 1] = (1.0, 0.0, 0.0);
+            b.Vertices[vidBase + 2] = (0.0, 1.0, 0.0);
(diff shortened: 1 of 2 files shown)


iamahsanmehmood (approved):
Verified locally: full .NET suite passes (69/69, including the new O(n) regression test - 100k instances in ~1s, well under the 30s ceiling), and the format/analyzers-as-errors build CI runs are clean. The deferred back-fill correctly reproduces the old scan's outermost-ancestor-wins semantics while fixing the O(instances x meshes) hot path, and also fixes a latent substring-match bug the old scan had (e.g. "A_B" matching inside "A_BC"). All required CI is green.

iamahsanmehmood:
Thanks for the fix and the detailed writeup — great catch on the O(n²) hot path, and the linear-time regression test is a nice guard against it coming back. Merged.

PR #190, opened 2026-08-18
feat(python): parse VFF scenes (pages) and linear dimensions

tuxiasumari:
Hi Ahsan — giving back some of what IngeTrazo's fork has been running in production since July: the two VFF entity families the Python reader doesn't surface yet.

## Scenes (pages)

The `0702` record nests `6D60 › 6D61 ›` one `7148` per page:

- `6F54 › 6F55` — page name (UTF-8)
- `714A › 34BC` — camera: `34BD`/`34BE`/`34BF` eye/target/up (3×f64, inches), `34C4` field of view, `34C2` u8 **perspective flag** (`00` = parallel projection), `34C3` f64 visible height when parallel
- `7150` — hidden layers as (u8 length, var-int layer id) runs

The projection flag and ortho height were calibrated against the scene thumbnails SketchUp embeds inside the `.skp` itself — parallel plans/elevations carry `00` and their `34C3` matches the thumbnail framing exactly, while perspective scenes carry `01` with a stale `34C3`.

Surfaces as `SkpModel.pages` (new `Page` dataclass); hidden layer ids resolve to layer names via the existing `layer_id_to_name` map.

## Linear dimensions

The `5BCC` record: `5BCD`/`5BCE` are the two connection points — type 1 is an explicit world-space point in `520A`; **type 2 is connected to geometry**: `520B › 53FC` names the vertex by persistent id (`53FD`) and the instance holding it (`53FE`, length-prefixed), so the definition-local vertex position is lifted to world by that instance's transform. `5BCF`/`5BD0` give the dimension plane, `5BD2` the offset. An unresolvable connection drops the whole dimension (fail-safe).

Two new per-record scans (`_scan_vertex_positions`, `_scan_instance_transforms`) accumulate the persistent-id maps the resolution needs — same streaming discipline as the rest of `full_parse`, nothing retained beyond two flat dicts.

Surfaces as `SkpModel.dimensions` with world-space endpoints in inches. The existing per-definition, text-only `Dimension` fields are untouched, so the legacy reader's output is unchanged (its dimensions simply have `a is None`).

## Validation

- Your own `tests/fixtures/Untitled.skp` turns out to carry **13 linear dimensions** — now parsed with full geometry, and the new tests pin them.
- On a 607 MB production survey file (7 scenes, mixed parallel/perspective, 5 hidden layers each), output is **identical** to the fork implementation IngeTrazo has been shipping, with no measurable parse-time cost (23.7 s vs 25.2 s on the same machine).
- Full suite: **309 passed, 29 skipped** (the optional-deps skips).

Classic (pre-2021) files import with `pages=[]` — I don't have a repro file with legacy scenes yet; the CPage walker can follow when one shows up.

### DIFF
--- a/packages/python/src/openskp/_core.py
+++ b/packages/python/src/openskp/_core.py
@@ -273,6 +273,279 @@ def find_child_tag(nodes, target):
             return res
     return None
 
+
+def _tlv_items(buf):
+    """Parse ``buf`` as a flat run of (u16 tag, u32 len, payload) records
+    covering the whole buffer, or ``None`` when it doesn't fit — the page
+    (scene) container nests plain TLV runs inside leaf payloads."""
+    items = []
+    off = 0
+    n = len(buf)
+    while off < n:
+        if off + 6 > n:
+            return None
+        tag = struct.unpack_from('<H', buf, off)[0]
+        ln = struct.unpack_from('<I', buf, off + 2)[0]
+        if tag == 0 or off + 6 + ln > n:
+            return None
+        items.append((tag, buf[off + 6:off + 6 + ln]))
+        off += 6 + ln
+    return items
+
+
+def _tlv_find(items, tag):
+    """First payload with ``tag`` in a ``_tlv_items`` list, or ``None``."""
+    for t, p in items or ():
+        if t == tag:
+            return p
+    return None
+
+
+def _scan_vertex_positions(top, id2pos):
+    """Accumulate every vertex's persistent id (hex) → ``(x, y, z)`` inches.
+
+    A vertex is a ``C409`` record: ``DC05`` holds its persistent id (the
+    ``DE05`` var-int payload), ``C509`` its 3×f64 position. Dimension
+    connection points reference geometry by this id (see
+    :func:`_parse_dimensions`). Ids are unique file-wide, so a single flat map
+    resolves a reference to the exact vertex regardless of which context holds
+    it — model-root dimensions reference model-root vertices (world space).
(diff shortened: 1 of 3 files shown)


iamahsanmehmood (approved):
Reviewed the actual byte-format logic, not just the description — this is solid:

- Both new families (pages/scenes via `0702`, linear dimensions via `5BCC`) are scoped correctly into the existing per-top-level-record streaming loop (`_scan_vertex_positions`/`_scan_instance_transforms`/`_find_page_node` all run inside the same `iter_top_level_lazy` loop `collect_defs` already uses) — keeps this project's core memory-bounded streaming architecture intact, doesn't reintroduce whole-file materialization.
- The dimension → instance-transform resolution (definition-local vertex lifted to world via the owning instance's composed matrix) is correct and has a sensible fail-safe (drops the dimension rather than guessing on an unresolvable reference) — verified with a synthetic test that checks both the success and drop paths.
- `Dimension`/`SkpModel` changes are purely additive with defaults — the existing legacy per-definition `Definition.dimensions` path is untouched, confirmed by reading both code paths.

Independently verified beyond CI: checked out the branch into a worktree, ran the new test file plus the full suite locally (338 passed), and ran `ruff check` — all clean. Manually inspected the resolved dimensions from `tests/fixtures/Untitled.skp` — 13 real dimensions with sane, non-degenerate coordinates and offsets, `model.pages` correctly empty for a file with no scenes.

Great addition — genuinely new capability (not just closing a parity gap), well-documented ground-truth calibration notes, validated against a real 607MB production file. Approving.

PR #192, opened 2026-08-19
fix(typescript): read v20 filler counts that are multiples of 256

thomazyujibaba:
## Summary

Follow-up to your review on #155, which flagged this:

> `retryCountAfterV20Filler`'s "skip zero padding, first non-zero byte is the count's low byte" heuristic has a narrow edge case: if the real count is an exact multiple of 256 (its own low byte is `0x00`), the scan can't distinguish that from filler padding and will misalign the read.

You were right. Fixing it.

## Type
- [x] Bug fix

## Changes

The padding length genuinely varies per call site — 9 and 13 bytes both occur in real files, so a fixed offset is not an option. But the count always lands at `markerAt + 4 + pad` with `pad % 4 === 1`, so the probe can step through those candidate offsets and read whole u32s instead of walking byte by byte. No individual byte is ever inspected, which is what made a `0x00` low byte ambiguous.

Two smaller things that came with it:

- **Each call site now passes the plausibility limit it already enforces** (`5_000_000` / `100_000` / `1_000_000`). The probe accepts exactly the range the caller would, rather than hardcoding one bound for all four.
- **The byte logic is split out as `findCountAfterV20Filler` and exported**, following `isClassRef`'s precedent, so the layout can be tested on synthetic records without a fixture.

## Testing

`tests/legacy-v20-filler.test.ts`, 7 cases. Three of them fail on the current implementation:

```
count    256 pad 13 | before: 4278190081 | after: 256
count    512 pad  9 | before: 4278190082 | after: 512
count  65536 pad 13 | before: 4294901761 | after: 65536
```

The other cases cover the paddings seen in real files, the resume offset, a non-empty string record (must be left alone), a missing marker, the per-caller limit, and a truncated buffer.

- 105/105 tests pass, `tsc --noEmit` and eslint clean
- All fixtures — v17, v20, VFF — produce identical results to before

On your note that this "mostly degrades to hitting the implausibility checks and throwing cleanly": that matches what I see, so this is hardening rather than a live bug fix. No file I have trips it.

### DIFF
--- a/packages/typescript/src/legacy.ts
+++ b/packages/typescript/src/legacy.ts
@@ -24,10 +24,60 @@ export class LegacyParseError extends Error {}
 
 const STR_MARKER = new Uint8Array([0xff, 0xfe, 0xff]);
 
+/** Widest zero padding seen between the v20 filler's empty string and the
+ * count that follows it (9 and 13 bytes occur in real files; the ceiling
+ * leaves room without letting the probe wander into unrelated records). */
+const MAX_V20_FILLER_PAD = 29;
+
 function bytesEqual(a: Uint8Array, b: Uint8Array): boolean {
   if (a.length !== b.length) return false;
   for (let i = 0; i < a.length; i++) if (a[i] !== b[i]) return false;
   return true;
+}
+
+/**
+ * Locates the count that follows a v20 filler record, given the offset the
+ * bad count was read from. Pure byte logic, exported for tests; see
+ * {@link retryCountAfterV20Filler} for how it is used.
+ *
+ * Returns the count and the offset just past it, or null when the bytes do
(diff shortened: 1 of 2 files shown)


iamahsanmehmood (approved):
Verified the algorithm myself: the ascending-pad probe order guarantees the first plausible u32 read is the true count, since every candidate before it is provably still inside the zero-padding run — so a count whose low byte is 0x00 (the exact case that broke the old byte-scan) can never be misread as padding. Tests cover the real padding lengths, the multiple-of-256 regression, the resume offset, a genuine non-empty string, a missing marker, the per-caller limit, and a truncated buffer. Clean, well-reasoned fix. Thanks for following up on the #155 review note.

PR #193, opened 2026-08-19
feat(typescript): embed texture images in GLB export (opt-in)

thomazyujibaba:
## Summary

The parser has extracted texture bytes for a while (#4, #13) and every baked primitive already carries UVs (#6), but `toGLB` never wrote `images` or `textures` — so an exported model loses its textures and every textured material falls back to its averaged colour. A wood-veneer cabinet exports as a flat brown block.

This wires up the last step. Since it changes the export's public shape and its size, it is **opt-in**.

## Type
- [x] New feature
- [x] Bug fix (the material-identity part below)

## Changes

**`toGLB(scene, { textures: true })` embeds the images** and points each textured material's `baseColorTexture` at them. Off by default because it is not free — a real fixture here goes 0.82 MB → 3.41 MB. With the option off the output is byte-identical to before, and `baseColorTexture` is stripped so no reference dangles into a `textures` array that isn't there.

**Material identity was also wrong, independently of the export.** `getMaterialIndex` deduplicated on colour alone, so two materials with *different images* but the same average RGB collapsed into one and one image was lost. Real files hit this: a store model here has two fabrics that both resolve to `141,141,141`. The texture is now part of the key, for both the material map and the face grouping. Primitive counts are unchanged (21 on the v17 fixture), so this does not fragment the mesh.

**`baseColorFactor` deliberately keeps the resolved colour** instead of being reset to white. glTF multiplies factor by texture, which is what SketchUp's colorized materials rely on — and zeroing it broke the existing `item 14 regression` test, which was right to complain.

**`SkpScene` gains `textures`**, deduplicated by image bytes. The mime type is sniffed from the magic bytes rather than the filename: SketchUp records the authoring machine's path (`H:/Drives compartilhados/.../Tauari.jpg`) and its extension can disagree with the content. A format glTF cannot carry (TIFF, from older files) is skipped and the material keeps its flat colour.

## Testing

`tests/glb-textures.test.ts`, 10 cases on `capilla_quiroz_v17.skp` (3 distinct JPEGs across 4 materials): collection and dedup, the default staying clean, the images really landing in the BIN chunk at 4-byte-aligned offsets, GLB structural validity both ways, and a no-op on an untextured model.

- 108/108 tests pass, `tsc --noEmit` and eslint clean
- Verified on real files by round-tripping the output through `@gltf-transform`: escrivaninha 0.82 → 3.41 MB with 7 textured materials, a store model 16.64 → 18.36 MB with 14, every image's bytes intact

## Notes for other platforms

Per CONTRIBUTING's parity note: Python, Dart, .NET and C++ have the same gap — none of their GLB writers emit `images`/`textures`, and their material dedup is colour-only too. The expected behavior is the above: texture as part of the material key, images deduplicated by bytes, mime sniffed from magic bytes, embedding gated behind a caller-supplied flag. I only have TypeScript set up here, so I have not ported it.

## Open question

I picked opt-in with the flag defaulting to off, so nobody's export changes size without asking. If you'd rather have it on by default (textures are arguably what most people expect from a GLB) or named differently, happy to change it — the mechanism is the same either way.

### DIFF
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -366,9 +366,49 @@ function createGlb(json: any, binaryBuffer: Uint8Array): Uint8Array {
  * @param scene - The result of buildScene()
  * @returns GLB file as Uint8Array
  */
-export function toGLB(scene: SkpScene): Uint8Array {
+/** Options for {@link toGLB}. */
+export interface GlbOptions {
+  /** Embed the scene's texture images in the GLB and point each textured
+   * material's `baseColorTexture` at them.
+   *
+   * Off by default because it is not free: a model with photographic
+   * textures can be several times larger with the images than without, and
+   * the geometry alone is what most callers are after. When off, textured
+   * materials fall back to their averaged base colour, which is what this
+   * exporter has always produced.
+   */
+  textures?: boolean;
+}
+
+/**
+ * Drops `baseColorTexture` from materials when the images are not being
+ * embedded. The reference would otherwise dangle - `textures[0]` would not
+ * exist - and a strict glTF reader rejects the file outright.
+ */
+function stripTextureRefs(materials: unknown[]): unknown[] {
+  let needsCopy = false;
+  for (const mat of materials) {
+    if ((mat as any)?.pbrMetallicRoughness?.baseColorTexture) {
+      needsCopy = true;
+      break;
+    }
+  }
+  if (!needsCopy) return materials;
+  return materials.map((mat) => {
+    const pbr = (mat as any)?.pbrMetallicRoughness;
+    if (!pbr?.baseColorTexture) return mat;
+    const restPbr = { ...pbr };
+    delete restPbr.baseColorTexture;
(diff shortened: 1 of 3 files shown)


iamahsanmehmood (approved):
Reviewed the diff and the material-identity fix in detail. The colour-only dedup bug is real and well-substantiated (two fabrics both averaging to 141,141,141 in a real file) - keying on texture index too is the right fix, and the added test explicitly guards against re-merging them. The opt-in gating is sound: verified stripTextureRefs only fires when no textures are embedded, so no dangling baseColorTexture reference can leak into a strict-reader-breaking file either way, and the default-off path stays byte-identical to before. Alignment, mime-sniffing-from-magic-bytes (right call given SketchUp stores the authoring machine's path, not a reliable extension), and the dedup-by-length+prefix heuristic are all reasonable, disclosed tradeoffs. Good coverage in glb-textures.test.ts. Appreciate the parity note for the other 4 languages - checklisting that for later.

PR #194, opened 2026-08-20
Fix five legacy-reader gaps found on real engineering models

tuxiasumari:
## Summary

Marco here (via Claude). While sweeping IngeTrazo's round-trip harness over a folder of real project files (2015-2026), two SketchUp models that open fine in real SketchUp were rejected by the legacy walker. Hunting those two errors uncovered five distinct reader gaps — all in `legacy.py`:

1. **Textured colour-by-layer** (`_read_layer`): a CLayer can embed the same texture payload a CMaterial carries. The reader assumed flat RGBA and walked into the JPEG bytes (`expected a string record`). The payload is now factored into `_texture_block()`, shared by both readers.
2. **CDimensionLinear's "165-byte opaque tail"** actually embeds two object references (connection points). They're 2 bytes in small archives but 6 once the archive passes 0x7FFE objects and the 0x7FFF big-tag escape kicks in — the fixed skip derailed exactly on large models (`back-ref to class slot 3`). Now read structurally (37 + ref + 42 + ref + 82).
3. **Text leader attachments & CRelationship pointers can reference forward** (the annotation serializes before the geometry it labels). A new `_entity_ref()` tolerates that where the strict back-ref path rightly does not.
4. **CImage had no reader** — any legacy file with a dropped image was rejected outright. The record is instance-shaped (back-ref to the already-walked definition holding the image's face/texture + 3×4 placement + path + GUID); calibrated byte-exact on two files.
5. **Attribute value type 0x11** (`Geom::Point3d`, 3 doubles) was unknown; extensions (road tools etc.) store point arrays with it. Same payload as the supported 0x12 Vector3d.

## Validation

- A real **661 MB SketchUp 2017 model** (the IngeTrazo calibration expediente) that the walker always rejected **now parses completely**: 85 definitions, 284,836 faces, 40 materials, 16 layers, 12.9 s.
- Byte-identical fingerprints (defs/faces/edges/verts/insts/materials/layers) on the previously-parsing legacy corpus (v16/v17/v18).
- Full test suite green: 320 passed, 30 skipped.

## Repro models (published for you to test)

https://github.com/tuxiasumari/openskp/releases/tag/repro-legacy-2026-08 — `plaza-toros-puquina.skp` (81 MB, v18) and `casa-bueno.skp` (17 MB, v18), real project files, both open fine in SketchUp. They exercise fixes 1-5 and also the one gap that remains (filed separately): the **annotation-connection sets** in definition tails, which still stop both files' full parse. The 661 MB file is available on request.

### DIFF
--- a/packages/python/src/openskp/legacy.py
+++ b/packages/python/src/openskp/legacy.py
@@ -429,12 +429,55 @@ def _read_layer(ar, r):
     while len(mid) < 8 and r.peek(3) != _STR_MARKER:
         mid += r.raw(1)              # flags: 3 bytes on v16, 4 on v17+
     r.utf16()                        # internal name ("Layer_<name>")
-    r.u16()
+    flags = r.u16()
+    if flags & 0x00FF:
+        # Colour-by-layer with a TEXTURED material: instead of the flat
+        # RGBA, the layer embeds the same texture block a CMaterial
+        # carries (SketchUp Pro assigns full materials to layers). Low
+        # byte of the flag word set = textured; a plain colour layer has
+        # 0 there (its high byte carries an unrelated flag, so the word
+        # as a whole is non-zero either way).
+        tex = _texture_block(ar, r)
+        r.raw(4)                     # trailing u32
+        return {'k': 'layer', 'name': name, 'hidden': mid[0] if mid else 0,
+                'rgba': tex['rgba']}
     rgba = r.raw(4)
     r.utf16()
     r.raw(21)
     return {'k': 'layer', 'name': name, 'hidden': mid[0] if mid else 0,
             'rgba': tuple(rgba)}
+
+
+def _texture_block(ar, r):
+    """The textured-material payload: an embedded CDib plus applied size,
+    source file name, average colour, and opacity. Shared verbatim between
+    a CMaterial with a texture and a colour-by-layer CLayer that carries a
+    textured material."""
+    r.raw(2 if ar.ver >= 17 else 1)     # texture flag pad
+    s, n, dib = ar.read_object(r, expect='CDib')
+    if not (isinstance(dib, dict) and dib.get('k') == 'dib'):
+        raise LegacyParseError(f"texture object is not a dib {r.ctx()}")
+    # optional u32 between the dib and the 2 x f64 applied size
+    marker = r.data.find(_STR_MARKER, r.pos, r.pos + 28)
+    if marker - r.pos == 20:
+        r.u32()
+    elif marker - r.pos != 16:
+        raise LegacyParseError(f"texture size block misaligned {r.ctx()}")
+    w = r.f64()
+    h = r.f64()
+    fname = r.utf16()
+    avg = r.raw(9)               # RGBA + 00 + RGBA (colour stored twice)
+    r.utf16()
+    blob = r.raw(8)              # u32 + u32 colorized flag
+    opacity = r.f64()
+    use_op = r.u8()
+    # A colourized (re-tinted) texture stores the ORIGINAL image plus
+    # the tint as the average colour; flagged by the second blob u32
+    # or by alpha 0xFF on the stored colour.
+    colorized = bool(blob[4]) or avg[3] == 0xFF
+    return {'rgba': tuple(avg[:4]), 'opacity': opacity, 'use_opacity': use_op,
+            'tex_dib': s, 'tex_w': w, 'tex_h': h, 'tex_file': fname,
+            'colorized': colorized}
 
 
 def _read_material(ar, r):
@@ -450,31 +493,7 @@ def _read_material(ar, r):
         use_op = r.u8()
         out.update(rgba=tuple(rgba), opacity=opacity, use_opacity=use_op)
     else:
-        r.raw(2 if ar.ver >= 17 else 1)     # texture flag pad
-        s, n, dib = ar.read_object(r, expect='CDib')
-        if not (isinstance(dib, dict) and dib.get('k') == 'dib'):
-            raise LegacyParseError(f"texture object is not a dib {r.ctx()}")
-        # optional u32 between the dib and the 2 x f64 applied size
-        marker = r.data.find(_STR_MARKER, r.pos, r.pos + 28)
-        if marker - r.pos == 20:
-            r.u32()
-        elif marker - r.pos != 16:
-            raise LegacyParseError(f"texture size block misaligned {r.ctx()}")
-        w = r.f64()
-        h = r.f64()
-        fname = r.utf16()
-        avg = r.raw(9)               # RGBA + 00 + RGBA (colour stored twice)
-        r.utf16()
-        blob = r.raw(8)              # u32 + u32 colorized flag
-        opacity = r.f64()
-        use_op = r.u8()
-        # A colourized (re-tinted) texture stores the ORIGINAL image plus
-        # the tint as the average colour; flagged by the second blob u32
-        # or by alpha 0xFF on the stored colour.
-        colorized = bool(blob[4]) or avg[3] == 0xFF
-        out.update(rgba=tuple(avg[:4]), opacity=opacity, use_opacity=use_op,
-                   tex_dib=s, tex_w=w, tex_h=h, tex_file=fname,
-                   colorized=colorized)
+        out.update(_texture_block(ar, r))
     return out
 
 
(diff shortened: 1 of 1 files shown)


tuxiasumari:
Update: the hunt is finished — **both repro models now parse completely**, so this PR (now 3 commits) closes #195 as well.

The "annotation-connection sets" theory in #195 turned out to be wrong in an instructive way: the mysterious block was ordinary CFace/CLoop/CEdgeUse geometry all along. The REAL mechanism was **burned store-map indices**: when a dimension or leader text anchors to geometry, SketchUp `MapObject`s its connection points into the MFC store map *without serializing any bytes*, so the writer's index counter advances and every later back-reference looks "forward" to a byte-exact reader. The fix turns the edge-use parent oracle into a resync signal instead of a hard failure, plus three smaller variants found on the way (drawbase layer-by-object — inline CLayer or escaped ref instead of the u16 id —, CConstructionLine's 4-byte v18 tail, and the earlier CImage/Point3d/dimension/text/layer fixes).

Validation:
- `casa-bueno.skp`: 200 defs, 56,884 faces, 83 materials — 2.4 s
- `plaza-toros-puquina.skp`: 521 defs, 259,267 faces, 1,575 instances, 430 materials, 36 layers — 13.4 s
- the 661 MB v17 expediente: unchanged, still parses completely
- previously-parsing corpus fingerprints byte-identical, suite green (320 passed)
- end-to-end through IngeTrazo's importer: both models load via the pure backend (9 s / 59 s), no skp2dae fallback needed

iamahsanmehmood:
Thanks for the deep dig here - fixes 1, 4, and 5 (textured colour-by-layer, `CImage`, and the `0x11` attribute type) all look correct on review, and the `_texture_block()` factoring is a nice cleanup. Fix 3 (`_entity_ref`, forward-tolerant refs on `CRelationship`/`CDimensionLinear`/`CText`) is currently low-risk since none of those new fields (`connect`, `attach`, `refs`) are consumed by `_fill_builder` yet - they're parsed but not wired in.

Fix 2, the `_resync_slots` mechanism, is where I want to slow down before merging. I reproduced it in isolation against your branch (not just read it) and found a real bug:

`_read_edgeuse` captures `es` (the edge's slot) *before* calling `_resync_slots`:

```python
es, _, _ = ar.read_object(r, expect='CEdge')
...
if 0 < delta <= 4096:
    _resync_slots(ar, delta)
...
return {'k': 'edgeuse', 'edge': es, 'sense': sense}
```

`_resync_slots` shifts every slot from `start` through `next_slot - 1` up by `delta`, which includes `es` itself whenever the edge was freshly allocated (the common case for the first edgeuse of a loop right after a burn - it's essentially *always* true for whichever edgeuse first detects the mismatch, since that's what triggers the resync). But `es` is a plain int already returned to the caller by that point, so nothing updates it. The old slot `es` isn't left empty either - because the shift runs as an in-place walk from high to low, a *different*, unrelated object from `delta` slots earlier ends up moved into `es`'s old position.

Concretely, with a synthetic repro (slots 0-99 filler, watermark=100, a `CLoop` at 150, a fresh `CEdge` allocated at 151, then a resync with delta=6 fired from that same edgeuse): the real `CEdge` ends up at 157, but `ar.slots.get(151)` - which is what `_fill_builder` looks up via the `es` still embedded in the returned `edgeuse` dict - now returns whatever object used to live at slot 145. `_add_edge` then either throws `KeyError: 'v1'` (if that object isn't edge-shaped) or, worse, silently attaches the *wrong* edge's vertices to this loop (if it happens to also be a `CEdge`, which is common in edge-dense geometry).

I think this is why it doesn't show up in your validation: the byte-identical fingerprint check runs against the previously-parsing v16/v17/v18 corpus, which by definition never exercises `_resync_slots` (those files parsed fine before this PR). The 661 MB file's "parses completely, plausible counts" is the only signal for this path, and a swapped/wrong edge reference doesn't change face/edge/vertex *counts* - only their topology - so it wouldn't show up that way even if it's happening.

The fix is probably to track the shift the same way `_new_obj` already does for its own slot (`epoch = ar.shift_total` before the read, then `+= ar.shift_total - epoch` after) - but applied to `es` specifically in `_read_edgeuse`, since `_new_obj`'s own correction only protects the edgeuse's *own* returned slot, not values already captured earlier in the same function body.

Happy to re-review once that's addressed - the rest of this PR is in good shape.

iamahsanmehmood (changes requested):
Requesting changes for the _resync_slots issue detailed in my comment above - a verified bug, not a stylistic concern. Everything else in this PR looks solid.

tuxiasumari:
Corpus report: we swept **186 real-project .skp files (2015-2026, up to 50 MB)** from Marco's archive through the branch. Result: **182/186 parse completely** (the VFF-era files all pass; the legacy set includes v16/v17/v18 with dimensions, leader texts, guide lines, images, colour-by-layer and extension attribute dictionaries).

The sweep surfaced one more one-liner, now in the branch (4th commit): attribute value type **0x0C = Length** (a plain double, inches) — four 2016 files store extension data with it.

The 4 remaining failures are two distinct undecoded variants, tracked with repro files for a future round:
- three 2016 "salinas" files: a misalignment near the layer-list end (`definition-list anchor is None` / `implausible definition count` — the anchor read lands inside a layer record, so some layer-tail variant precedes it);
- one 2017 file: a guide-line (CConstructionLine) trailer variant, same family as the v18 4-byte tail fixed earlier (the record carries a 16-byte GUID-like trailer here).

tuxiasumari:
Great catch — thank you for actually reproducing it rather than just reading. You were right on every count, including why our validation was blind to it (fingerprints count entities; they don't check topology).

Rather than patching `es` specifically, the 5th commit **reworks the mechanism so the failure class can't exist**: registrations now always stay at walker indices (nothing ever moves, so no captured slot integer — `es`, an inlined edge's `v1`/`v2`, anything — can go stale), and `_backref` translates FILE references through the recorded burn bands (piecewise offset per event). References INTO a band — the phantom connection points — resolve to an explicit `'reserved'` placeholder instead of whatever object a shift would have left at that index.

Nice side effect that confirms your diagnosis: on `casa-bueno.skp` the new mechanism removes exactly one bogus edge the shifting version let through (118,073 → 118,072 edges — a phantom-band ref that used to resolve to a real, wrong edge and got appended as a list item). Everything else is byte-identical across the corpus; both repro models, the 661 MB file, and the 182/186 sweep still pass; suite green (320). Your synthetic scenario (filler 0-99, loop at 150, fresh edge at 151, delta=6) is structurally impossible now — the edge registers at 151 and stays there; the file's later refs to it (157) translate back to 151.

Also noted on fix 3 being parse-only for now — agreed, `connect`/`attach`/`refs` are deliberately not wired into `_fill_builder` yet; they're there so the stream walks correctly and the data is preserved for whenever dimensions get surfaced.

iamahsanmehmood (approved):
The rework addresses the _resync_slots bug directly: registrations now stay at walker indices permanently (nothing moves, so no captured slot like `es` can go stale), and file-side back-references are translated through the recorded burn bands instead. Traced through the new-object vs. back-ref ordering by hand and it holds up - the burn-detecting oracle (the loop-parent ref) fires on every edgeuse, so any shared edge downstream of a burn will already have that burn registered by the time it's referenced a second time.

Validation is thorough: 182/186 real-project corpus files now parse, byte-identical fingerprints on the previously-passing corpus, and the new mechanism visibly fixes a real wrong-edge bug the old shifting version was silently letting through on casa-bueno.skp. CI green across all five languages.

Approving.

tuxiasumari:
Final corpus result with the 6th commit: **186/186 — the entire 2015-2026 real-project corpus parses completely.** The last four failures were two more decoded variants (v20 layer-list null separators + the filler landing after a definition's decl with a zero-low-byte count, and per-writing-build guide-line tails, now self-calibrated per file). One upstream fixture assertion was corrected along the way: `gondola_v20.skp` really carries a second layer ("Gondulas Laterais") that the old counting dropped — the test had enshrined the bug.

All six files that ever failed have also been verified visually in IngeTrazo against SketchUp Web side by side. This closes out everything we can find to hunt; the branch is stable from our side whenever you want to re-review.

iamahsanmehmood:
Since this PR was already merged by the time your 6th-commit comment landed, pushing further to `fix-legacy-bugs` here didn't actually reach `main` - GitHub doesn't pull new commits into a PR's target branch once it's closed. Cherry-picked it (your authorship preserved) into a fresh PR and merged it: #199, now in `main` at `17d0fcb`. Corpus is 186/186 - nice work chasing the last four down. If you find anything else, either a new PR from a fresh branch or just pinging here works; a push to an already-merged branch won't land on its own.

PR #199, opened 2026-08-20
Decode the last legacy-corpus failures: v20 layer separators and per-build guide-line tails

iamahsanmehmood:
## Summary

Marco's follow-up to #194 — he kept sweeping the 186-file real-project corpus after that PR was already merged and pushed a 6th commit to his fork's `fix-legacy-bugs` branch, but since the PR was closed by then it never landed. Cherry-picked here (author preserved) to get it into `main`.

Closes out the last remaining corpus failures — the three 2020 "salinas" files and the 2017 "Iglesia la Barrera Final" file, all in `packages/python/src/openskp/legacy.py`:

1. **v20 layer lists** (model-level and per-definition) interleave a null object-ref after each layer record; the declared count is the number of real layers, not items. The old counting walked off mid-list on any v20 file with several layers (surfaced as "definition-list anchor is None" / "implausible definition count"). Now counts parsed layers, skips the separators, and stops early if a back-ref (the anchor) arrives.

2. **v20 can drop its undocumented filler right after a definition's decl field**, swallowing the u32 that normally follows the count. The filler helper now also runs there (v20 only), and the count anchor is hardened to try up to three earlier positions since the count's low byte can legitimately be zero.

3. **`CConstructionLine`'s trailing block varies by SketchUp's writing build**, not cleanly by version (7 bytes on the v17 calibration corpus, 4 on v16 and a real v18, 0 on another real v17). Now self-calibrated per archive on the file's first guide line instead of assumed from the version number.

Also corrects a real bug in the existing v20 fixture test: `gondola_v20.skp` genuinely carries a second layer ("Gondulas Laterais") that the old counting silently dropped — the test had enshrined the bug as correct behavior.

## Validation (from Marco's PR #194 comment)

- Corpus now **186/186** (up from 182/186 after #194) — the 3 "salinas" project files (309/192/186 defs, up to 113,875 faces and 3,396 instances) and the "Iglesia la Barrera Final" file (295 defs, 42,820 faces, 123 layers) all parse completely.
- Previously-passing corpus fingerprint-identical.
- All six files that ever failed also verified visually in IngeTrazo against SketchUp Web, side by side.

Independently re-run before opening this PR: `pytest` (350 passed) and `ruff check src/ tests/` (clean) on this exact branch.

Co-authored-by: Marco Sumari Tellez <ing.sumari@gmail.com>

### DIFF
--- a/packages/python/src/openskp/legacy.py
+++ b/packages/python/src/openskp/legacy.py
@@ -242,7 +242,22 @@ class _Archive:
         return slot, ent[1], ent[2]
 
 
-def _retry_count_after_v20_filler(r: _R, count_pos: int) -> Optional[int]:
+def _plausible_list_tag(ar, data, at) -> bool:
+    """True when the u16 at ``at`` can legally start an object read: a
+    null, an escape, a class definition, a class-ref to a KNOWN class, or
+    an object back-ref within the allocated range."""
+    if at + 2 > len(data):
+        return False
+    t = struct.unpack_from('<H', data, at)[0]
+    if t in (0x0000, 0x7FFF, 0xFFFF):
+        return True
+    if t & 0x8000:
+        ent = ar.slots.get(t & 0x7FFF)
+        return ent is not None and ent[0] == 'class'
+    return t < ar.next_slot
+
+
+def _retry_count_after_v20_filler(r: _R, count_pos: int, ar=None) -> Optional[int]:
     """SketchUp 2020 (v20) writes an extra, undocumented record ahead of some
     counts that v17 does not have, which leaves the reader a few bytes early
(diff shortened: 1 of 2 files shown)


iamahsanmehmood:
Reviewed the diff directly against Marco's own explanation in #194's final comment - all three mechanisms (v20 layer-list null separators, the v20 post-decl filler drop, and the per-build CConstructionLine tail) match what he described, and the gondola_v20.skp fixture fix is a genuine correction (the old count really was silently dropping a real layer). Re-ran pytest (350 passed) and ruff independently on this exact branch before opening the PR. Merging with --admin since I opened this PR myself (cherry-picked Marco's commit off his fork after his original PR had already merged and closed) - can't self-approve.

PR #200, opened 2026-08-21
feat(typescript): add instancing-preserving scene output (buildInstancedScene + toInstancedGLB)

thomazyujibaba:
## Summary

Adds an **instancing-preserving scene output** to the TypeScript package: `buildInstancedScene()` plus a dedicated `toInstancedGLB()` exporter.

`buildScene()` bakes every placed component/group instance into its own world-space vertex and index buffers, so scene size and peak memory scale with

```
definition geometry × number of placed instances
```

rather than

```
unique geometry + instance transforms
```

A component placed 1,000 times costs 1,000 copies of its `Float32Array` and index buffers. Consumers that want the file's original instancing back are left to recover it heuristically — recentering baked primitives, comparing geometry, and guessing which placements shared a definition. Since the `.skp` already recorded that instancing, the information is being discarded and then reconstructed.

The new API preserves it directly. **This is lossless instancing preservation, not mesh decimation:** no vertices are removed, merged, quantised or approximated, and no new runtime dependencies are introduced. The triangles are exactly the ones `buildScene()` produces — stored once and referenced N times instead of copied N times.

`parseSkp()`, `buildScene()` and `toGLB()` are **unchanged** in behaviour, return type and output. TypeScript only; the Python, .NET, Dart and C++ ports are untouched.

## Type
- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### Public API

```ts
buildInstancedScene(buffer, options?): InstancedScene   // + SkpFile#buildInstancedScene()
toInstancedGLB(scene, options?): Uint8Array
```

The result separates the three things cleanly:

- `meshResources` — reusable **local-space** mesh resources (`InstancedMeshResource[]`, each with `LocalPrimitive[]`)
- `sceneHierarchy` — placed instance nodes with transforms and metadata (`InstancedNode`)
- `gltfMaterials` / `textures` — materials and textures, same shape as `SkpScene`

I kept the names from the issue description; they read consistently with the existing `buildScene`/`toGLB` pair, so there was no reason to invent a different convention. No conditional return type was added to `buildScene()`.

### Resource identity is *not* the definition ID

The same definition renders differently depending on where it is placed, so caching on definition ID alone would silently repaint geometry. A resource variant is keyed on the **effective rendering context**:

- the inherited instance material (SketchUp's "paint the component") — which sets both unpainted-face colour and, through its texture's tile size, their UVs;
- **texture identity**, not averaged colour — two different images can average to the same RGB (real files do this);
- the effective **layer's fallback colour**, which is what an entirely unpainted face renders as.

Instances with the same effective context share one resource; those that differ get separate variants. Front/back material resolution, per-face UV mapping and double-sided-vs-split geometry follow deterministically from the definition plus those inputs, so they need no separate key component. `InstancedMeshResource.variantKey` exposes the resolved context. IDs (`mesh_0`, `mesh_1`, …) and ordering are deterministic.

### Geometry fidelity is structural, not reimplemented

Rather than duplicating the face-processing logic, I extracted the existing **local-space** face grouping verbatim into `src/face-groups.ts` and made both builders call it — triangulation (including faces with holes), UV seams and positioned textures, front/back materials, double-sided handling, per-(vertex, uv) splitting.

This was a small, contained extraction, not a rewrite of `buildSceneFromParsed()`: that function's structure, error handling, progress reporting and output are untouched, and the whole existing suite passed unchanged immediately after the extraction, before any new code was added. The only difference between the two paths is what happens *after* grouping — the baked path applies the instance matrix to the vertices, the instanced path leaves them local and puts the transform on the node.

### Transform correctness

- Instance matrices convert to glTF space as the similarity transform `C·M·C⁻¹` with `C: (x,y,z) → (x,z,-y)`, which **composes correctly through nesting**: converting each level and multiplying equals converting the fully composed matrix.
- **Normals stay local.** Normal transformation is deferred to the node transform (glTF's own rule: inverse-transpose of the upper-left 3×3). That is precisely what keeps non-uniform and mirrored (negative-determinant) scales correct *without* baking a per-placement copy of the normal buffer.
- Nothing is baked to make these cases pass — the tests below run the mirrored and non-uniform cases through both paths and compare.

### Coordinates and units (documented explicitly)

SketchUp stores inches, Z-up. Both builders convert to **metres, glTF Y-up**. `LocalPrimitive.positions`/`normals` are metres/Y-up in **definition-local** space; `InstancedNode.matrix` is a 16-element **column-major, parent-relative** glTF matrix. The one deliberate exception is `positionMm`, kept as **absolute millimetres on SketchUp Z-up** so it matches the baked path's `InstanceNode.positionMm` for metadata comparison.

### GLB export

`toInstancedGLB()` emits valid glTF 2.0/GLB where multiple nodes reference the same mesh; binary vertex/index buffers are written **once per resource**, not per instance. Hierarchy, transforms, materials, textures, UVs and names are preserved. A definition resolving to several materials becomes **one mesh with several primitives** (glTF's normal representation) rather than extra nodes. `{ textures: true }` embeds images, matching `toGLB()`'s opt-in.

## Testing

`npm test` → **247 passed, 31 files** (203 pre-existing, all still passing unchanged + 44 new).

The strongest evidence is `tests/instanced-fixture-parity.test.ts`, which runs **both** builders over the repository's real `.skp` fixtures — `SU_File.skp`, `Untitled.skp`, `capilla_quiroz_v17.skp`, `gondola_v20.skp`, `single_material_v17.skp`, covering both the modern VFF and legacy MFC v17/v20 containers — flattens the instanced result back to world space with a **test-only** helper, and asserts equality of triangles and resolved materials, plus identical per-node layer/properties/position metadata.

Two comparison details worth flagging, since both looked like bugs at first and neither was:

- **Materials are compared by content, not index.** Both paths build the same material table but allocate into it in their own encounter order (the baked path during its flattened walk, the instanced path as each resource is first created). The tables are identical as sets; only the indices differ.
- **Positions are compared with a tolerance of 1e-5 m**, justified by float32 precision rather than fitted to make the suite pass. The baked path transforms in float64 then stores world-space float32; the instanced path stores local-space float32 and transforms afterwards — same geometry, different rounding point. Measured worst-case delta across all fixtures is **~1.6e-6 m on a ~47 m model**, which is about one float32 ulp at that magnitude. Comparison is in walk order, not nearest-neighbour: these models contain many coincident/mirrored duplicate triangles, and a nearest-match strategy pairs the wrong twin and reports phantom differences.

The flatten helper is deliberately test-only — materialising the flattened form is the exact thing the feature exists to avoid.

Covering the requested cases, in `tests/instanced-scene.test.ts` and `tests/instanced-glb.test.ts`:

| # | Requirement | Test |
|---|---|---|
| 1 | One definition, many placements → one resource, many nodes | `emits ONE mesh resource for a definition placed many times` |
| 2 | Instance count does not grow resource buffers | `keeps mesh-resource buffer bytes flat as instance count grows` (asserts **equality**, not sub-linearity) |
| 3 | Nested transforms match `buildScene()` | `matches for NESTED instance transforms` |
| 4 | Identical inherited materials share | `shares ONE resource when the effective inherited material matches` |
| 5 | Different inherited materials split | `splits into SEPARATE resource variants...` |
| 6 | Layer fallback colour in identity | `includes the layer fallback colour in resource identity` (+ the converse: two layers sharing a colour *do* share) |
| 7 | UV seams / textures preserved | `preserves UVs from the inherited textured material`, `deduplicates textures shared by several materials` |
| 8 | Front/back materials correct | `matches for different FRONT and BACK materials`, `marks a face whose sides resolve alike as doubleSided, once` |
| 9 | Root loose geometry retained | `retains ROOT loose geometry` |
| 10 | Mirrored / non-uniform scale | `matches for a MIRRORED (negative-determinant) placement`, `matches for uniform and non-uniform scale` |
| 11 | GLB reuses one mesh from many nodes | `reuses ONE glTF mesh from many nodes`, `does NOT duplicate binary buffers per instance` |
| 12 | Existing regressions unchanged | full suite green; plus `existing toGLB output is unchanged` |

Also covered: faces with holes, recursive-definition rejection (matching the baked path), accessor/bufferView consistency against the binary chunk, and valid instanced GLB round-trips on the real fixtures.

Per the request, **no timing assertions** were added to CI — tests assert structural scaling only; timings live in the benchmark.

### Verification commands

All run in `packages/typescript`, matching `ci-typescript.yml` (typecheck → build → test, and the separate lint job):

```
$ npm run typecheck    # tsc --noEmit — clean
$ npm run build        # tsup — ESM 259.10 KB, CJS 261.63 KB, DTS 58.65 KB, success
$ npm test             # Test Files 31 passed (31) / Tests 247 passed (247)
$ npm run lint         # eslint src/ — clean
```

I also verified the **built** package surface (`dist/index.js` + `dist/index.d.ts`) exports `buildInstancedScene`, `toInstancedGLB`, `SkpFile#buildInstancedScene`, and the `InstancedScene` / `InstancedMeshResource` / `InstancedNode` / `LocalPrimitive` types, and round-trips a real fixture.

Note: `npx eslint tests/` reports 7 pre-existing errors in `create/ifc/json-export/legacy-v20/parser` test files, untouched by this PR (CI lints `src/` only). My new test files lint clean. I left those alone as unrelated.

## Benchmark

Added `npm run bench:instanced` (`tests/bench/instanced-bench.ts`) — deterministic synthetic scene, one 24-face component repeated N times, median of 5 runs, kept out of the test suite so CI stays stable.

**Measured on Node v22.20.0, win32/x64** (real output, not estimated):

| N | `buildScene` ms | `buildInstancedScene` ms | baked buf KB | instanced buf KB | baked prims | mesh resources | instance nodes | baked GLB KB | instanced GLB KB |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 | 0.37 | 0.46 | 3.6 | 3.6 | 1 | 1 | 1 | 4.6 | 4.7 |
| 10 | 3.06 | 0.23 | 35.6 | 3.6 | 10 | 1 | 10 | 43.3 | 5.4 |
| 100 | 15.92 | 0.47 | 356.3 | 3.6 | 100 | 1 | 100 | 432.6 | 12.8 |
| 1000 | 146.61 | 4.50 | 3562.5 | **3.6** | 1000 | **1** | 1000 | 4338.7 | 90.0 |

At N=1000: geometry buffers **1000× smaller**, GLB **48× smaller**, build **~33× faster**. Instanced buffer bytes are *identical* across all four N — the geometry is stored exactly once. At N=1 the instanced path is marginally larger/slower, as expected: there is nothing to share, and it pays for node structure.

(Timings vary between runs on a loaded machine — an earlier run of the same script showed 338 ms vs 7.40 ms at N=1000, i.e. ~46×. The byte columns are exact and reproducible.)

## Documentation

`packages/typescript/README.md` gains a **Choosing an API** section covering when to use `parseSkp()` vs `buildScene()` vs `buildInstancedScene()`, the memory tradeoff (with the scaling formulas and measured numbers), coordinate systems and units, how material variants affect resource reuse, and a worked instanced-GLB export example. It states explicitly that this is lossless instancing preservation, not mesh decimation. Also updated: the feature list, Quick Start, Exporting snippet, and the module table. A `[Unreleased]` entry was added to the root `CHANGELOG.md` following the existing format.

## Checklist
- [x] Tests pass locally (247/247)
- [x] Documentation updated

## Notes for the reviewer

- **Two-tier API.** `buildInstancedScene()` re-parses independently, consistent with `parseSkp()`/`buildScene()`, so callers who never ask for it never pay for it. I did not restructure parsing to share a single pass — that would have meant touching the existing entry points, and compatibility mattered more here.
- **`vite-node` added as an explicit devDependency.** It was already present transitively via `vitest`, and the benchmark script invokes it directly; relying on a transitive dep felt fragile. Lockfile change is 3 lines. No runtime dependencies added.
- **Hidden-state behaviour is matched, not "fixed".** `buildScene()` currently does not filter hidden faces/instances/layers out of the bake; the instanced path deliberately does the same, so the two agree. Changing that is a separate behavioural decision.
- The diff is the feature only — no generated files, build output, fixtures, or unrelated edits.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,37 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added — TypeScript
+
+**Instancing-preserving scene output.** A new `buildInstancedScene()` keeps
+the instancing SketchUp already recorded, instead of baking it out the way
+`buildScene()` does. Each distinct definition is triangulated once, in its
+own local space, and every placement becomes a node carrying the transform
+that puts it there — so output scales with `unique geometry + instance
+transforms` rather than `definition geometry x placement count`. A
+companion `toInstancedGLB()` writes glTF 2.0/GLB in which many nodes
+reference one mesh, so a definition's vertex and index buffers are written
+once rather than once per placement.
+
+This is lossless instancing preservation, not mesh decimation: no
+vertices are removed, merged, quantised or approximated, and no new
+dependencies are introduced. The triangles are exactly the ones
+`buildScene()` produces. The test suite asserts that directly, by
+flattening the instanced result and comparing it against the baked one on
+the repository's real `.skp` fixtures.
+
+On a synthetic scene of one 24-face component repeated 1,000 times, the
+geometry buffers are 1,000x smaller (3,562 KB to 3.6 KB), the exported GLB
+is 48x smaller, and the build is 46x faster. Run
+`npm run bench:instanced` in `packages/typescript` to reproduce.
+
+`parseSkp()`, `buildScene()` and `toGLB()` are unchanged in behaviour,
+return type and output. TypeScript only; the Python, .NET, Dart and C++
+ports are untouched. See
+[Choosing an API](packages/typescript/README.md#choosing-an-api).
+
 ## [1.1.0] — 2026-08-20
 
 **Write support, now in all five languages.** OpenSKP can *create* new
--- a/packages/typescript/README.md
+++ b/packages/typescript/README.md
@@ -42,6 +42,11 @@ SDK (see [Writing](#writing) below).
   call.
 - **Scene baking** — an opt-in `buildScene()` pass resolves the full placed
   scene graph to world-space, triangulated, export-ready geometry.
+- **Instancing preserved** — `buildInstancedScene()` keeps SketchUp's own
+  instancing instead of baking it out: unique geometry once, plus one
+  transform per placement, so a component placed 1,000 times costs one copy
+  of its buffers. Losslessly — no decimation or quantisation. See
+  [Choosing an API](#choosing-an-api).
 - **Native multi-format conversion** — glTF (GLB), Wavefront OBJ/MTL, STL,
   PLY, AutoCAD DXF (3DFACE and Polyface Mesh), IFC4 (BIM/ISO 10303-21 STEP),
   and JSON — all written from scratch, no third-party CAD/BIM SDK involved.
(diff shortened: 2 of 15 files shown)


iamahsanmehmood (approved):
Reviewed in depth, not just the write-up:

- face-groups.ts extraction is genuinely verbatim - diffed the old inline logic in model.ts against the new shared function line-by-line, including the subtle inheritedMaterial?.color ?? fallbackLayerColor fallback rule.
- Re-derived the C*M*C^-1 coordinate-conversion math by hand rather than trusting the description - correct, and consistent with the row-major convention already used by the existing (untouched) transformPoint().
- Normal handling matches the baked path's zero-length fallback (1e-6 threshold -> [0,0,1]); deferring the transform to the node matrix is what keeps mirrored/non-uniform scales correct without a per-instance buffer copy.
- GLB buffer sharing confirmed structurally in instanced-glb.ts: binary buffers are written once per resource, node-walk only ever sets gltfNode.mesh - valid glTF instancing.
- The fixture parity test isn't tautological: asserts triangle-count equality before the ordered comparison (so a dropped-triangle bug can't hide behind a min-length comparison), compares materials by resolved content rather than index, and the 1e-5m tolerance is justified by float32 rounding-point math across two different paths, not picked to make the suite pass. Runs on 5 real fixtures across both VFF and legacy MFC v17/v20.

CI green across all 5 languages' required checks. Only TypeScript changed. One cosmetic-only nit (package.json's description field em-dash got re-escaped to \u2014 by whatever tool touched it) - not worth blocking on.

Approving.

Issue #201, opened 2026-08-21
[FEATURE] TypeScript: cut geometry memory with flat typed arrays (measured 4.6x on vertices, 23x on edge flags)


thomazyujibaba:
## Summary

Four concrete, independent memory optimisations for the **TypeScript** geometry storage, aimed squarely at the limitation the README already documents:

> Memory use scales worse than the other four ports on very large files — a 113 MB file needs 8–16 GB of Node heap […] Root cause: V8's per-object overhead on millions of small geometry objects.

I measured that root cause rather than assuming it. The overhead is real and it is concentrated in a handful of `Map`-of-boxed-values structures in `GeometryBuilder` and in the eager object materialisation in `buildModelFromParsed()`. Replacing those with flat typed arrays is a **4.6x** reduction on vertices and **23x** on edge flags, with no change to parsing logic or output.

Opening this for discussion before writing any code — item 1 in particular touches a structure that the parser, the triangulator and both scene builders all read, so I would rather agree on the approach (and on whether it is wanted at all) than arrive with a large unsolicited diff.

Not proposing decimation, quantisation, Draco, meshopt, WASM or any lossy operation anywhere in this issue. Every item below is a pure representation change.

Context: I recently opened #200 (instancing-preserving scene output), and profiling for that work is what surfaced these. They are independent of #200 and of each other.

## Motivation

The five ports agree on the parsing architecture (one top-level record at a time, per `DEVELOPER_GUIDE.md#performance`), and that fixed the outright crashes. What remains in TypeScript is not orchestration but **representation**: each `[x, y, z]` is a separate JS array with its own object header, and each `Map` entry carries bucket overhead on top. That is what puts TS at 8–16 GB where Dart needs a 4 GB heap flag and .NET needs nothing.

### Measured overhead

Node v22.20.0, win32/x64, `--expose-gc`, RSS delta after forced GC, values retained so nothing is optimised away:

| Structure | Current representation | Flat typed array | Ratio |
|---|---|---|---|
| 1M vertices | `Map<number, [x,y,z]>` — **111.1 MB** | `Float64Array` — 24.4 MB (22.9 theoretical) | **4.6x** |
| 1M edge flags | `Map<number, number>` — **29.6 MB** | `Uint8Array` — 1.3 MB (1.0 theoretical) | **23x** |

Reproduce:

```js
// node --expose-gc
const N = 1_000_000, keep = [];
[...]

PR #202, opened 2026-08-21
perf(typescript): store edge display flags in a Uint8Array, not a Map (10.3x on real fixtures)

thomazyujibaba:
## Summary

Item 2 from #201, as greenlit: `GeometryBuilder.edgeFlags` moves from `Map<number, number>` to a `Uint8Array`-backed store.

The payload is a **single byte** per edge (D307 display flags: base `0x06`, plus `0x01` hidden, `0x08` soft, `0x10` smooth). A `Map` entry costs roughly 30 bytes once V8's boxed-number and hash-bucket overhead are counted — about 30x the data itself.

## Type
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [x] Performance / internal representation

## Changes

New `src/edge-flags.ts` exporting `EdgeFlagStore`: a `Uint8Array` indexed by `id - baseId`, grown geometrically, and **re-based** when an id arrives below the current base (nothing guarantees ascending id order, so this can't assume it).

The diff to existing files is 6 lines — one field declaration and one import in each of `geometry.ts` and `legacy.ts`. All the complexity is contained in the new module.

### Answering the question you raised: are edge ids dense?

You asked for a real-file number rather than the synthetic one, since sparse ids would mean either wasting memory sized to the max id or clawing back the gain with an id→index indirection. Measured across every fixture in the repo:

| Fixture | Edges | Id span | Density | `Map` | `Uint8Array`+bitmap | Gain |
|---|---:|---:|---:|---:|---:|---:|
| `SU_File.skp` | 84 | 343 | 24.5% | 2.5 KB | 0.4 KB | 6.5x |
| `Untitled.skp` | 12,762 | 26,551 | 48.1% | 373.9 KB | 29.2 KB | 12.8x |
| `capilla_quiroz_v17.skp` | 515 | 2,275 | 22.6% | 15.1 KB | 2.5 KB | 6.0x |
| `gondola_v20.skp` | 9,186 | 29,069 | 31.6% | 269.1 KB | 31.9 KB | 8.4x |
| **Aggregate** | **22,547** | **58,238** | **38.7%** | **660.6 KB** | **64.0 KB** | **10.3x** |

**Conclusion: ids are sparse (~39%), and straight span-indexing still wins by 10.3x — so no indirection is needed here.** One `Uint8Array` slot costs one byte even when unused, versus a whole hash entry per stored edge, so the Map only breaks even below roughly 3% density, which no real file approaches.

I'd rather put the honest 10.3x on the record than the 23x from the synthetic dense benchmark in the issue. The 23x is what you'd get with dense ids; 10.3x is what this actually delivers.

(Same measurement, incidentally, gives the vertex-id density number you wanted before design-approving item 1 — 32.6% aggregate, worst single definition 7.6%. Posted in full on #201; item 1 will need the id→index mapping as you predicted, and its real gain is 2.1–3.9x rather than 4.6x.)

### Preserving a distinction the Map made

The legacy (pre-2021 MFC) reader only records an edge when its flags are **non-zero**:

```ts
const flags = (db.soft ? 0x08 : 0) | (db.smooth ? 0x10 : 0) | (db.hidden ? 0x01 : 0);
if (flags) {
  builder.edgeFlags.set(slot, flags);
}
```

So "absent" and "stored zero" are genuinely different states, and a bare `Uint8Array` (where an untouched slot reads 0) would conflate them and change parsed output. `EdgeFlagStore` keeps a separate presence bitmap — 1 bit per slot, already counted in the table above — so `get()` returns `undefined` for an id never written, exactly as `Map.prototype.get` does. Callers' existing `?? 0` fallbacks behave identically.

## Testing

`npm test` → **256 passed, 32 files** (247 pre-existing all still passing, unchanged + 9 new).

The three existing `parser.test.ts` assertions on `builder.edgeFlags.get(...)` pass **unchanged** — `get`/`set`/`has`/`size` keep Map semantics, so the store is drop-in.

New `tests/edge-flags.test.ts` covers the cases where a hand-rolled typed-array store realistically goes wrong:

- growth far beyond the first id written;
- **re-basing** when an id arrives below the current base, including repeated re-basing interleaved with growth (the bitmap can't be block-copied on a shift, since it isn't byte-aligned to it — that's the subtle one);
- **stored zero vs absent**, per the legacy behaviour above;
- overwrite not double-counting `size`;
- byte masking;
- a deterministic randomised 2,000-write pattern asserted key-by-key against a reference `Map`, so any divergence in semantics fails reproducibly (seeded LCG, no `Math.random`).

Real-file coverage comes for free: both `.skp` container formats route through this field, and the existing legacy/VFF fixture suites plus the `buildScene`/`buildInstancedScene` parity tests all exercise it.

### Verification

```
$ npm run typecheck    # tsc --noEmit — clean
$ npm run build        # tsup — ESM/CJS/DTS success
$ npm test             # Test Files 32 passed (32) / Tests 256 passed (256)
$ npm run lint         # eslint src/ — clean
```

## Checklist
- [x] Tests pass locally (256/256)
- [x] Documentation updated (n/a — internal representation, no public API change)

## Notes

- **No public API change.** `GeometryBuilder` is imported into `index.ts` but never re-exported, as you confirmed; `Edge.soft`/`smooth`/`hidden` are still derived exactly as before in `buildDefinition()`. Published consumer types are untouched.
- **No new dependencies**, and nothing lossy — this is representation only, same bytes in and out.
- Scope is `packages/typescript`; the other four ports are untouched.
- Sequenced per your suggestion: this first, item 1 once you've looked at the density number, item 3 last (and I'd understand deferring it given the Worker interaction).

### DIFF
--- a/packages/typescript/src/edge-flags.ts
+++ b/packages/typescript/src/edge-flags.ts
@@ -1,0 +1,111 @@
+/**
+ * Storage for edges' D307 display-flag byte, keyed by edge id.
+ *
+ * The payload is a SINGLE BYTE per edge (base 0x06, plus 0x01 hidden,
+ * 0x08 soft, 0x10 smooth), which a `Map<number, number>` stores at roughly
+ * 30 bytes per entry - about 23x the data itself, once V8's boxed-number
+ * and hash-bucket overhead is counted. On a large model that is tens of
+ * megabytes spent on a value that fits in a `Uint8Array` slot.
+ *
+ * Backed by a `Uint8Array` indexed by `id - baseId`, grown geometrically.
+ * Edge ids are not dense in real files (~39% across this repository's
+ * fixtures), so the array is sized to the observed id SPAN rather than the
+ * edge count - still far smaller than the Map it replaces, because one
+ * slot costs one byte instead of a whole hash entry.
+ *
+ * Semantics match the `Map` exactly, including the distinction the legacy
+ * (pre-2021 MFC) reader relies on: it only records an edge when its flags
+ * are non-zero, so an id that was never written must read back as 0 and
+ * report `has() === false`. A separate presence bitmap keeps that
+ * observable difference intact rather than conflating "absent" with
+ * "stored zero".
+ */
+export class EdgeFlagStore {
+  /** Flag byte per slot; index is `id - baseId`. */
+  private flags: Uint8Array = new Uint8Array(0);
+  /** One bit per slot recording whether that id was ever written. */
+  private present: Uint8Array = new Uint8Array(0);
+  /** Id that maps to slot 0. Set on first write. */
+  private baseId = 0;
+  private initialized = false;
+  private count = 0;
+
+  /** Number of ids actually stored. */
+  get size(): number {
+    return this.count;
+  }
+
+  private slotFor(id: number): number {
+    return id - this.baseId;
+  }
+
+  /** Grow (and, for an id below `baseId`, shift) so `id` has a slot. */
+  private ensureSlot(id: number): number {
+    if (!this.initialized) {
+      this.baseId = id;
+      this.flags = new Uint8Array(64);
+      this.present = new Uint8Array(8);
+      this.initialized = true;
+      return 0;
+    }
+
+    if (id < this.baseId) {
+      // An id below the current base: re-base, shifting existing data up.
+      const shift = this.baseId - id;
+      const needed = shift + this.flags.length;
+      const nextFlags = new Uint8Array(Math.max(needed, this.flags.length * 2));
+      nextFlags.set(this.flags, shift);
+      this.flags = nextFlags;
+
+      const nextPresent = new Uint8Array(Math.ceil(nextFlags.length / 8));
+      // Re-stamp presence bit by bit: the bitmap is not byte-aligned to the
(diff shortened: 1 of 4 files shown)


iamahsanmehmood (approved):
Reviewed EdgeFlagStore directly, not just the description - traced the re-basing path by hand (the trickiest part: growing/shifting the presence bitmap when an id arrives below the current base). The loop bound after reassignment looked suspicious at first (this.flags.length is read AFTER this.flags is reassigned to the new, bigger array), but working through both branches of the max() confirms it can never under-iterate - worst case it does a few harmless extra checks against out-of-bounds reads, which JS typed arrays return as undefined (coerces to 0 in the bitwise check), never data loss. Confirmed against the actual read site in model.ts:391 (?? 0 fallback) that get() returning undefined for absent matches Map semantics exactly, so this is a genuine drop-in.

Appreciate the honesty on the numbers - correcting the issue's synthetic 23x down to the real-file 10.3x, and stating the vertex-density finding for item 1 even though it wasn't asked for in this PR specifically.

CI green, no public API surface change (confirmed GeometryBuilder is still never re-exported), no new dependencies. Approving.

PR #203, opened 2026-08-21
Write and read annotations in legacy .skp: linear dimensions and leader texts

tuxiasumari:
Follow-up to #194/#199, adding **annotation support to both sides of the legacy pipeline**: `create.SkpBuilder` gains `add_dimension()` and `add_text()`, and the reader now surfaces dimension geometry and text anchors on the model (`SkpModel.dimensions`, `TextEntity.point` / `label_point`).

## The method: SDK-generated ground truth ("Rosetta" files)

Rather than guessing at record layouts from human files (arbitrary floats, hard to align), we built two MinGW helpers that drive the **real SketchUpAPI.dll** under Wine to create entities with controlled, greppable values and save as SU2017 — then harvested the byte-exact records. Notes with the full decoded layouts are in `docs/dimension-record-notes.md`. Highlights:

- **CDimensionLinear**: the B37/B42 blocks are *connection blocks* — `[flags][u32 type][u32 4][point3d]` with type 1 = free point inline, 2 = anchored back-ref (same enum as VFF). The free-dimension placement head is a constant; the offset patches in at B82+62.
- **CText**: the SDK itself can only author *screen* texts (its two doubles after the font are screen fractions — SketchUp renders them superimposed at view centre, detached from the model; leader vectors never serialize, even anchored via `SUInstancePath`). Real leader texts were decoded from human-drawn corpus files instead: screen slot zeroed, free-connection anchor, the label's world position in the placement tail, and a leader-type u32 (2 = pushpin) before the arrow delimiter. `add_text()` writes that shape.

## Reader fixes the ground truth exposed

- `_read_text`'s delimiter predicate hard-required arrow type 3, silently dropping any text drawn with a different arrow (the root-list tolerance ate the error).
- `_read_skfont` consumed the pid mask but not the pid bytes it declares.
- Dimensions/texts now resolve their endpoints (explicit or via entity refs) and reach the model — the `connect`/`refs` data #194 preserved is finally wired into `_fill_builder`.

## Validation

- Every generated file passes the official SDK (skp2dae accepts) **and** was verified visually in SketchUp Web: dimensions in X/Y/Z/diagonal orientations render with correct values; leader texts hang anchored with their leader lines. Round-trips through our own reader are exact.
- Upstream suite green (320 passed), and the 186-file real-project corpus from #194/#199 stays byte-identical on fingerprints (casa bueno's own human-drawn leader text now yields both its anchor and label position).

Consumed in production by IngeTrazo's .skp exporter/importer (dimensions and leader texts now survive the trip both ways between IngeTrazo and SketchUp).

### DIFF
--- a/docs/dimension-record-notes.md
+++ b/docs/dimension-record-notes.md
@@ -89,6 +89,22 @@ A second SDK helper (`mktext`) settled the text record. Key findings:
 - `_read_skfont` consumed the pid MASK but not the pid bytes it
   declares — fixed; SDK-written fonts also carry a longer tail, which
   the text reader's delimiter scan absorbs.
-- Leader VECTORS did not persist through `SUTextSetLeaderVector` in the
-  harness (records carry only the anchor point); point-labels are what
-  `add_text()` writes, which matches IngeTrazo's text entities.
+- Leader VECTORS never persist through the SDK — not even when the
+  text is anchored to real geometry via `SUInstancePath` (a second
+  harness, `mktext2`, proved it: face- and vertex-anchored records
+  save fine but the leader stays zeroed, and face+leader texts are
+  silently DROPPED from the file). The SDK can only author SCREEN
+  texts: the two doubles after the font are SCREEN FRACTIONS (0.5,
+  0.5 = view centre) and SketchUp renders every such text superimposed
+  there, detached from the model.
+- Human-drawn LEADER texts (casa bueno, quiroz) settle the real
+  grammar. Connection types: 1 = free 3D anchor inline, 2 = anchored
+  to an entity back-ref (like dimensions), 7 = attached to a FACE
+  (two u,v bbox fractions + face ref). The placement tail after the
+  connection is `[12B zeros][point3d LABEL world position][16B zeros]
+  [f64 1.0][u32 LEADER TYPE (0 none / 2 pushpin)][delimiter]` — the
+  leader line joins the label position to the anchor.
+- `add_text()` therefore writes the HUMAN shape, not the SDK's:
+  screen slot zeroed, free connection with the anchor, label position
+  at ``point + leader``, leader type 2, closed arrow (3). The reader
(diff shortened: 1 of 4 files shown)


iamahsanmehmood (approved):
Deep-reviewed the reader diff directly against the writer diff, not just the narrative - specifically cross-checked the connection-block byte offsets each side assumes:

- Connection 1 (b37): writer emits type at offset 5, point at offset 13 (bytes(5) + u32(1) + u32(4) + 3 doubles). Reader reads type at struct.unpack_from('<I', b37, 5) and point at offset 13. Exact match, zero slack to the block's 37-byte length.
- Connection 2 (b42): writer emits type at offset 10, point at offset 18. Reader matches exactly, zero slack to 42 bytes.

That kind of exact agreement between independently-read writer and reader code (not just "the tests pass") is what gives me confidence this is genuinely byte-correct, not just self-consistent by luck.

Also traced the _fill_builder dimension-endpoint resolution (free point takes precedence over anchored-vertex lookup, both must resolve or the dimension is dropped rather than emitting a partial one) and confirmed the unguarded dm["a"]/dm["b"] indexing in model.py's shared (pre-existing, VFF-original) conversion loop is safe specifically because full_parse_legacy's own filter (if d.get('a') and d.get('b')) guarantees those keys exist before they reach it.

The iterative commit history - especially catching its own wrong B82 assumption via the real SDK rejecting a computed value outright, and the two real reader bugs (_read_skfont's pid-byte count, the text-delimiter arrow-type over-restriction) found along the way - is exactly the kind of ground-truth-driven rigor this project's legacy-format work has needed throughout.

One trivial, non-blocking note: model.py's pre-existing "# Convert model-level linear dimensions (VFF; world space)" comment is now slightly stale since this same shared loop also handles legacy-sourced dimensions after this PR - not worth a follow-up on its own, just flagging.

CI green across all 5 languages (only Python changed). Approving.

Issue #204, opened 2026-08-21
[FEATURE] TypeScript: opt-in respectEdgeVisibility — soft/smooth/hidden flags are parsed but never acted on


thomazyujibaba:
## Summary

Split out from #201 as you suggested, so it can be judged on its own merits.

`parseSkp()` already exposes `Edge.soft`, `Edge.smooth`, `Edge.hidden` and `Face.hidden`, decoded correctly from the D307 display-flag byte. **Nothing downstream consults them.** `buildScene()` bakes all geometry regardless, and `toDXF`/`toOBJ`/`toPLY`/`toSTL` emit everything they are given.

Proposal: an opt-in flag on the scene builders — `buildScene(buffer, { respectEdgeVisibility: true })` and the same on `buildInstancedScene()` — that skips geometry SketchUp itself would not draw. Default `false`, so existing behaviour and every current test are untouched.

I want to be precise about what this is and isn't worth, because I overstated it in passing on #201 and the real numbers are more nuanced.

## Motivation

There is an asymmetry worth naming: `create.ts` **writes** these flags (`softEdges`, `smoothEdges`, `hiddenEdges`, `hidden` on faces and instances), and the parsers **read** them, but no consumer path ever **acts** on them. OpenSKP can author a model with hidden geometry and can tell you an edge is hidden — but every export it produces draws that edge anyway. Round-tripping a file through OpenSKP silently loses the visibility distinction it was careful to record on both ends.

### Measured impact — and it varies a lot by model

Edges by flag across the repository's fixtures:

| Fixture | Edges | soft | smooth | hidden | Drawable | Extra if ignored |
|---|---:|---:|---:|---:|---:|---:|
| `SU_File.skp` | 84 | 28.6% | 28.6% | 0.0% | 71.4% | 1.4x |
| `Untitled.skp` | 12,762 | 0.2% | 0.2% | 0.0% | **99.8%** | 1.0x |
| `capilla_quiroz_v17.skp` | 515 | 0.0% | 0.0% | 5.6% | 94.4% | 1.1x |
| `gondola_v20.skp` | 9,186 | 24.5% | 24.5% | 41.6% | **33.9%** | **3.0x** |
| **Aggregate** | 22,547 | | | | 72.7% | **1.4x** |

[...]

PR #205, opened 2026-08-22
feat(typescript): opt-in visibility filtering — isDrawableEdge + respectEdgeVisibility

thomazyujibaba:
Implements #204. Two of your three suggestions landed as proposed; the third didn't survive contact with the source, and I'd rather flag that than quietly ship something that doesn't do what the PR claims.

## One correction: `toDXF()` can't act on edge flags

You suggested `toDXF()` as the primary target, since it's "a real edge-consuming exporter that would show the actual 3x." I checked before implementing — it isn't one:

```ts
export function toDXF(scene: SkpScene, scale = METRES_TO_INCHES, mode: '3dface' | 'polyface' = 'polyface')
```

It takes a baked `SkpScene` and reads `scene.glbPrimitives`, emitting 3DFACE / Polyface Mesh entities built from **face triangles**. It never sees an edge, so it has no flags to consult — the `POLYLINE`/`VERTEX` records it writes are Polyface Mesh structure, not SketchUp edges.

That holds for every exporter: `toDXF`, `toOBJ`, `toPLY`, `toSTL` and `toIFC` all take `SkpScene`, and `grep '\.edges' src/{dxf,obj,ply,stl,ifc}.ts` returns nothing. **There is no edge-consuming path inside the library today** — the only edge consumers are downstream users reading `parseSkp()` output.

So an opt-in on `toDXF()` would have been the same "correctly implemented but decorative" problem you correctly identified for the scene builders, just relocated. Adding one would have meant either threading raw edges into an exporter that's built around baked scenes, or emitting a different entity type — both much larger than this issue warrants, and neither obviously right.

Your point stands, though; it just lands on item 2 instead. The 3x case is real and it belongs to consumers, so the helper is doing the load-bearing work here.

## What's in the PR

**1. `isDrawableEdge(edge)` — exported, documented.** For consumers working directly with `parseSkp()` output. `edge.soft || edge.smooth || edge.hidden` isn't self-evidently "SketchUp doesn't draw this" unless you know the format, so the intent gets a name.

```ts
const model = parseSkp(buffer);
const visible = model.root.edges.filter(isDrawableEdge);
```

Takes a structural type (`Pick<Edge, 'soft'|'smooth'|'hidden'>`), so it works on anything edge-shaped.

**2. `SceneOptions.respectEdgeVisibility` on both scene builders — for hidden faces.** Per your item 3: real and correct, and the docs say plainly that it won't move the aggregate number. Both builders share `face-groups.ts`, so it's one filter site, and the #200 parity test guards that they stay in agreement. On the instanced path a hidden definition's **node survives** with no `meshResourceId` — geometry drops, hierarchy and metadata stay intact.

**3. README section** covering both, with the honest per-model spread rather than the headline.

Off by default throughout. No behaviour, return-type or output change.

## Measurements

Non-drawable edges, per fixture:

| Fixture | Edges | Non-drawable | Effect |
|---|---:|---:|---:|
| `Untitled.skp` | 12,762 | 0.2% | ~1.0x |
| `capilla_quiroz_v17.skp` | 515 | 5.6% | ~1.1x |
| `SU_File.skp` | 84 | 28.6% | ~1.4x |
| `gondola_v20.skp` | 9,186 | **66.1%** | **~3.0x** |
| **Aggregate** | 22,547 | **27.3%** | ~1.4x |

Hidden faces: **0** across every fixture — which is why item 3's honest framing matters, and why there's a test asserting the option changes nothing on real files today (it'll notice if a future fixture does carry one).

## Testing

`npm test` → **264 passed, 33 files** (256 pre-existing unchanged + 8 new).

`tests/edge-visibility.test.ts` covers: the helper's four flag combinations; filtering real fixture edges without mutating the model; the near-no-op case on a flat model (documenting that the win isn't universal); default-off behaviour; hidden faces dropping when enabled on both builders; both builders agreeing on triangle count when enabled; and the option being a no-op on real fixtures.

```
$ npm run typecheck   # clean
$ npm test            # 264 passed (33 files)
$ npm run lint        # clean
```

## Checklist
- [x] Tests pass locally (264/264)
- [x] Documentation updated

## Notes

- `packages/typescript` only; other four ports untouched.
- Not proposing a default change or parse-time filtering — the flags stay exposed raw so a consumer can decide differently.
- If you'd still like edge-level output from an exporter, that's a genuinely bigger design question (which entity type, and whether an exporter should take `SkpModel` alongside `SkpScene`) and I'd rather scope it separately than bolt it on here.

### DIFF
--- a/packages/typescript/README.md
+++ b/packages/typescript/README.md
@@ -199,6 +199,43 @@ resolved context if you need to see why a definition produced more than one
 resource. Resource IDs (`mesh_0`, `mesh_1`, …) are stable and deterministic
 for a given file.
 
+### Edge and face visibility
+
+SketchUp does not draw every edge it stores. Three flags suppress an edge:
+`hidden` (explicitly hidden), and `soft`/`smooth` — the smoothing flags
+that make a faceted surface read as curved. That last pair is why a
+rounded model carries far more edges than it appears to: every curve is
+triangles stitched by edges that define the shape and are never shown.
+
+These are parsed and exposed on `Edge`, but acting on them is opt-in.
+
+```typescript
+import { parseSkp, isDrawableEdge } from 'openskp';
+
+const model = parseSkp(buffer);
+const visible = model.root.edges.filter(isDrawableEdge);
+```
+
+How much that saves is strongly model-dependent — measured across this
+repository's fixtures, 27.3% of edges are non-drawable on aggregate, but
+that ranges from **0.2% on a mostly-flat model to 66.1% on a
+curved-surface one**. Use it in wireframe/hidden-line renderers built on
+`parseSkp()` output, where drawing suppressed edges is both slower and
+visually wrong.
+
+For faces, both scene builders take an opt-in flag:
+
+```typescript
+const scene = buildScene(buffer, { respectEdgeVisibility: true });
+```
+
+This skips faces carrying SketchUp's "Hide" flag. It does **not** filter
+edges, because neither scene builder emits edges — their output is face
+triangles. Hidden faces are rare in practice (none in this repository's
+fixtures), so expect this to be correct rather than dramatic; the edge
+helper above is where the real saving lives. Off by default, since what
+SketchUp draws is a display policy rather than a parsing fact.
+
 ### Exporting an instanced GLB
 
 ```typescript
--- a/packages/typescript/src/face-groups.ts
+++ b/packages/typescript/src/face-groups.ts
@@ -44,6 +44,10 @@ export interface FaceGroupContext {
   fallbackLayerColor: { r: number; g: number; b: number };
   /** Identifies the definition in a triangulation failure. */
   definitionId: number | string;
+  /** Skip faces SketchUp itself would not draw (see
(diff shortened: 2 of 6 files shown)


iamahsanmehmood (approved):
Good catch correcting my own suggestion rather than just implementing it as stated - I verified it myself before approving: grepped all 5 exporters for .edges and got zero matches, confirmed toDXF() reads scene.glbPrimitives (baked face triangles) same as the others. My original toDXF suggestion was wrong; isDrawableEdge as the primary mechanism is the right call.

Reviewed the actual wiring: respectVisibility threads through face-groups.ts identically for both buildSceneFromParsed and buildInstancedSceneFromParsed (confirmed in the diff), isDrawableEdge takes a structural Pick<Edge,...> type as described, and the docs are upfront that the scene-builder option only covers hidden faces, not edges - no overclaiming. The instanced path's "node survives with no meshResourceId" behavior reuses the existing empty-primitives-returns-undefined path from #200 rather than introducing new logic.

CI green, no public API breakage, off by default throughout. Approving.

PR #206, opened 2026-08-22
perf(typescript): write 16-bit glTF indices where they fit (−7.8% GLB, −12.6% instanced)

thomazyujibaba:
## Summary

Both GLB exporters wrote every index as `UNSIGNED_INT` (4 bytes) regardless of primitive size. glTF 2.0 also allows `UNSIGNED_SHORT`, which every renderer supports and GPUs prefer — the index buffer is the most frequently-fetched buffer in a draw call, so halving it halves that bandwidth.

Real primitives sit far below the 65,536-vertex limit. Across this repository's fixtures **all 266 primitives qualify**, the largest being 2,742 vertices, so in practice the index data always halves.

Purely an encoding choice at the export boundary: in-memory `GlbPrimitive.indices` / `LocalPrimitive.indices` stay `Uint32Array`, so **no public type changes** and no consumer has to care. The other exporters (OBJ/STL/PLY/DXF/IFC) read the same in-memory arrays and are untouched.

## Type
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [x] Performance

## Changes

New `src/gltf-indices.ts` with `encodeIndices()`, used by both `toGLB()` and `toInstancedGLB()`.

Narrowing is keyed on the **largest index value**, not the vertex count — correct even for indices that don't start at zero — and falls back to `UNSIGNED_INT` with the original array untouched (no extra allocation) when anything exceeds the limit.

### The alignment consequence, handled explicitly

This is the real risk in the change, and it's why the diff touches the layout pass as well as the write pass. A 16-bit index buffer of **odd length** leaves the running byte offset 2-byte aligned. The next primitive's `POSITION` accessor is float32, which glTF requires to be 4-byte aligned — so writing 16-bit indices naively produces a spec-invalid file that many loaders still happen to accept.

Both exporters now pad to 4 bytes after each index buffer, in the size-computation pass and the write pass alike, so the two stay in agreement.

## Measurements

| Fixture | `toGLB` before → after | `toInstancedGLB` before → after |
|---|---|---|
| `SU_File.skp` | 7.5 → 6.9 KB (−8.1%) | 7.6 → 6.9 KB (−8.1%) |
| `Untitled.skp` | 717.7 → 607.7 KB (−15.3%) | 721.6 → 611.6 KB (−15.2%) |
| `capilla_quiroz_v17.skp` | 67.1 → 61.6 KB (−8.2%) | 61.8 → 56.7 KB (−8.3%) |
| `gondola_v20.skp` | 5023.0 → 4686.0 KB (−6.7%) | 298.5 → 277.1 KB (−7.2%) |
| **Total** | **5815.3 → 5362.2 KB (−7.8%)** | **1089.4 → 952.3 KB (−12.6%)** |

The percentage varies with the ratio of index data to vertex data — it's larger on the instanced exporter precisely because instancing already removed the duplicated vertex buffers, leaving indices a bigger share of what's left.

## Testing

`npm test` → **274 passed, 33 files** (256 pre-existing, all unchanged + 18 new).

### Validated against the official Khronos validator

Rather than relying only on my own assertions about spec compliance, I ran both exporters' output through `gltf-validator@2.0.0-dev.3.10`, on all four fixtures, with textures embedded:

```
VALID  SU_File.skp.baked.glb              err=0 warn=0
VALID  SU_File.skp.inst.glb               err=0 warn=0
VALID  Untitled.skp.baked.glb             err=0 warn=0
VALID  Untitled.skp.inst.glb              err=0 warn=0
VALID  capilla_quiroz_v17.skp.baked.glb   err=0 warn=0
VALID  capilla_quiroz_v17.skp.inst.glb    err=0 warn=0
VALID  gondola_v20.skp.baked.glb          err=0 warn=0
VALID  gondola_v20.skp.inst.glb           err=0 warn=0
```

8/8 files, **zero errors and zero warnings**. Happy to add this as a dev-dependency-backed test if you'd want it in CI, though it pulls a fairly heavy package for one check.

`tests/gltf-indices.test.ts` covers the encoder in isolation (boundary values at exactly 65535, keying on value rather than length, empty input, fallback above the limit) and then, per fixture and per exporter: that indices really are emitted as `UNSIGNED_SHORT`, that **every accessor's absolute offset is a multiple of its component width** (the spec rule the padding exists to satisfy), and that index values **round-trip byte-exactly** when read back out of the binary chunk — so narrowing provably doesn't renumber or reorder anything.

```
$ npm run typecheck   # clean
$ npm test            # 274 passed (33 files)
$ npm run lint        # clean
```

## Checklist
- [x] Tests pass locally (274/274)
- [x] Documentation updated (README note on the GLB exporters)

## Notes

- Branches off `main` and is independent of #205 (no file conflicts); the test count above is against `main` without it.
- **Two related ideas I measured and am NOT proposing**, since the data didn't support them:
  - *Deduplicating vertices shared between primitives of one definition.* Looked promising (692 of 1,548 vertices in one `gondola` resource share a position) but only **32** are identical in position **and** normal **and** UV. The rest legitimately differ per-face and glTF requires them separate. Worse, the primitives they span carry different materials, and an indexed vertex belongs to exactly one primitive — so there is essentially nothing to merge.
  - *Interleaving attributes into one buffer.* Doesn't change byte count at all; it would halve the bufferView count (104 → 52 on `gondola`, shrinking a JSON chunk that's 13% of the file) and improve GPU cache locality. Real but modest, and it changes the exporter's output layout more invasively than this. Worth its own issue if you're interested; I didn't want to bundle it here.
- Scope is `packages/typescript`; other four ports untouched.

### DIFF
--- a/packages/typescript/README.md
+++ b/packages/typescript/README.md
@@ -225,7 +225,9 @@ what they always have.
 ```typescript
 import { toGLB, toInstancedGLB, toOBJ, toMTL, exportOBJ, toSTLAscii, toSTLBinary, exportSTL, toPLYAscii, toPLYBinary, exportPLY, toDXF, exportDXF, toIFC, exportIFC, toJSON } from 'openskp';
 
-// Serialize a built scene straight to .glb bytes (in-memory, no disk I/O)
+// Serialize a built scene straight to .glb bytes (in-memory, no disk I/O).
+// Index buffers are written as UNSIGNED_SHORT when every index fits, which
+// is the usual case - roughly halving the index data at no loss.
 const glbBytes = toGLB(scene);
 
 // Instancing-preserving GLB: one mesh, many nodes (see Choosing an API)
--- a/packages/typescript/src/gltf-indices.ts
+++ b/packages/typescript/src/gltf-indices.ts
@@ -1,0 +1,64 @@
+/**
+ * glTF index-buffer narrowing.
+ *
+ * Both GLB exporters previously wrote every index as UNSIGNED_INT (4
+ * bytes), regardless of how small the primitive was. glTF 2.0 also allows
+ * UNSIGNED_SHORT, and every renderer supports it - it is in fact what GPUs
+ * prefer, since a 16-bit index buffer halves the bandwidth of the most
+ * frequently-fetched buffer in a draw call.
+ *
+ * Real primitives sit far below the 65,536-vertex limit: across this
+ * repository's fixtures the largest is 2,742 vertices, so every primitive
+ * qualifies and the index data halves.
+ *
+ * This is purely an encoding choice at the export boundary. The in-memory
+ * `GlbPrimitive.indices` / `LocalPrimitive.indices` stay `Uint32Array`, so
+ * no public type changes and no consumer has to care.
+ */
+
+/** Largest vertex index addressable by UNSIGNED_SHORT. */
+export const UINT16_INDEX_LIMIT = 65535;
+
+/** glTF `componentType` for UNSIGNED_SHORT. */
+export const COMPONENT_TYPE_UNSIGNED_SHORT = 5123;
+/** glTF `componentType` for UNSIGNED_INT. */
+export const COMPONENT_TYPE_UNSIGNED_INT = 5125;
+
+/**
+ * The narrowest glTF index encoding that represents `indices` losslessly.
+ *
+ * Keyed on the largest VALUE present, not on the primitive's vertex count:
+ * narrowing is only safe if every index actually stored fits, and keying on
+ * the value is correct even if a caller hands over indices that don't start
+ * at zero.
+ *
+ * Returns the componentType alongside a view ready to copy into the binary
+ * chunk. When narrowing isn't possible the original array is returned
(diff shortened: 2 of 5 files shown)


iamahsanmehmood (approved):
Reviewed the alignment-padding logic specifically, since that's the real risk here: traced both the size-computation pass and the write pass in toGLB and toInstancedGLB by hand and confirmed they apply the identical (4 - (x % 4)) % 4 padding operation, in the same order, to two counters that both start at 0 and accumulate the same byte sequence - so they're guaranteed to stay in lockstep by construction. Confirmed encodeIndices() only narrows after checking max <= 65535, so the Uint16Array truncation on narrowing is always lossless. Confirmed bufferView.byteLength correctly reports the unpadded data length (padding is just unused trailing bytes, which is valid glTF), not the padded total.

The Khronos gltf-validator run (8/8 files, zero errors/zero warnings) is stronger external evidence than I could produce by reading code alone, and the two "measured but not proposing" ideas (vertex dedup, buffer interleaving) show real investigation rather than padding the PR with easy wins.

CI green, no public API change (in-memory arrays stay Uint32Array), independent of #205. Approving.

PR #207, opened 2026-08-22
feat(typescript): expose embedded thumbnails and scene bounds

thomazyujibaba:
## Summary

Two things an asset browser or block library needs from a `.skp` — both of which the library already had internally but never surfaced, so every consumer rebuilt them.

**`extractThumbnail(buffer)`** returns the preview image SketchUp already saved inside the file. A catalogue can show what a model contains without parsing geometry, running a renderer, or generating previews offline.

**`SkpScene.bounds` / `InstancedScene.bounds`** give the model's axis-aligned extent, with `size` and `center` alongside `min`/`max` — answering "how big is this block?" and "where do I point the camera?", which previously meant sweeping every position buffer by hand.

## Type
- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### Thumbnails

```ts
const thumb = extractThumbnail(buffer);   // or SkpFile#thumbnail()
// { data: Uint8Array, mimeType, width, height, source: 'model' | 'preview' }
```

Reads container metadata only — no geometry parsing.

**Which stored image is returned is not arbitrary.** SketchUp writes two, and they are visibly different:

| Entry | Content |
|---|---|
| `meta/model_thumbnail.png` | The model on a clean background — **preferred** |
| `meta/preview_thumbnail.png` | The same view **with the red/green/blue axis lines drawn in** |

The axes read as clutter on a product card, so the model image wins and `preview` is only a fallback for a file that lacks it. `source` reports which was used. Both are 256×256 PNGs in the fixtures here.

### Returning `null` instead of guessing

`extractThumbnail()` returns `null` rather than throwing when there is no usable preview. That covers two real cases, and the second is a deliberate limitation worth calling out:

- **Modern files with no thumbnail entry** (e.g. written by a tool other than SketchUp).
- **Legacy (pre-2021 MFC) files.** These *do* embed PNGs — `gondola_v20.skp` carries three 256×256 ones, `capilla_quiroz_v17.skp` likewise — but the container stores images **without entry names**, so a thumbnail cannot be distinguished from a material's texture image without heuristics. I'd rather return `null` than hand back a texture labelled as a preview. (`Untitled.skp`, a VFF file, has exactly this trap: a `materials/_/thumbnail.jpg` sitting alongside the real ones.)

A malformed container still throws `SkpParseError`, matching the rest of the library.

### Bounds

```ts
const scene = buildScene(buffer);
if (scene.bounds) {
  const [w, h, d] = scene.bounds.size;   // metres, glTF Y-up
  scene.bounds.center;                    // e.g. to frame a camera
}
```

`null` when the scene has no geometry, so an empty model stays distinguishable from one sitting at the origin.

Two implementation details that are load-bearing rather than incidental:

**The baked path reads values back out of the `Float32Array`** rather than using the float64 intermediates it just computed. Bounds must describe the vertex data *as stored*, so that a consumer sweeping `positions` itself gets the identical answer. Using the float64 values is arguably *more* accurate but disagrees by a float32 ulp — which the test caught (1.4e-6 m on a 46 m model).

**The instanced path can't do the same**, since its geometry stays in local space. It walks the node tree and transforms each resource's local box by the accumulated matrix — only the **8 corners**, since an affine transform maps a box's corners to the corners of the transformed box, so the axis-aligned result is exact at a fraction of the cost of transforming every vertex. Reporting the union of local boxes instead would be meaningless as a model size; a test asserts both builders agree per fixture.

Accumulated during the bake, so it costs no extra pass over the vertex data.

## Testing

`npm test` → **303 passed, 35 files** (282 pre-existing, all unchanged + 21 new). Rebased onto `main` after #205 and #206 merged.

`tests/thumbnail-bounds.test.ts` covers:

- thumbnails extracted from both VFF fixtures, verified as real PNGs by signature (not just "something parsed"), 256×256;
- the `model`-over-`preview` preference, which is observable because both entries exist in those files;
- `null` for all four legacy fixtures;
- throwing on a non-SketchUp file;
- that extraction is cheaper than a full parse (deliberately loose — an order-of-magnitude property, not a benchmark, so it can't turn flaky in CI);
- reachability via `SkpFile`;
- bounds matching a manual sweep of the baked positions, per fixture, to 6 decimal places;
- baked and instanced bounds agreeing per fixture (1e-5 m, the same float32 tolerance the existing parity tests use);
- `null` bounds on `blank_v17.skp`, which has no geometry.

```
$ npm run typecheck   # clean
$ npm test            # 303 passed (35 files)
$ npm run lint        # clean
```

Also verified against the **built** package (`dist/index.js`): `extractThumbnail`, `SkpFile#thumbnail`, and `scene.bounds` all resolve, and the extracted PNG opens as a clean 256×256 cover image with no axis lines.

## Checklist
- [x] Tests pass locally (303/303)
- [x] Documentation updated (new "Cataloguing models" README section)

## Notes

- `bounds` is a **new field** on the scene types, so existing consumers are unaffected; `parseSkp()`, `buildScene()`, `toGLB()` and `toInstancedGLB()` are otherwise untouched. No new dependencies.
- Scope is `packages/typescript`; the other four ports are untouched. If the thumbnail idea is welcome, the VFF entry names are the same across ports, so it would port straightforwardly — happy to leave that to whoever owns each.
- Rebased onto `main` now that #205 and #206 have merged; the README and one `instanced.ts` import needed manual resolution, and both prior sections are preserved intact.
- If you'd prefer the legacy case to make a best-effort guess rather than return `null`, say so and I'll add it behind an explicit opt-in — I just didn't want guessing to be the default behaviour.

### DIFF
--- a/packages/typescript/README.md
+++ b/packages/typescript/README.md
@@ -236,6 +236,45 @@ fixtures), so expect this to be correct rather than dramatic; the edge
 helper above is where the real saving lives. Off by default, since what
 SketchUp draws is a display policy rather than a parsing fact.
 
+### Cataloguing models
+
+Two things an asset browser or block library needs, without paying for a
+full parse or a render:
+
+```typescript
+import { extractThumbnail, buildScene } from 'openskp';
+
+// The preview image SketchUp already saved inside the file. Reads
+// container metadata only - no geometry parsing, no renderer.
+const thumb = extractThumbnail(buffer);
+if (thumb) {
+  // thumb.data (raw bytes), thumb.mimeType, thumb.width, thumb.height
+  fs.writeFileSync('cover.png', thumb.data);
+}
+
+// The model's overall size, computed during the bake.
+const scene = buildScene(buffer);
+if (scene.bounds) {
+  const [w, h, d] = scene.bounds.size;       // metres, glTF Y-up
+  console.log(`${w.toFixed(2)} x ${h.toFixed(2)} x ${d.toFixed(2)} m`);
+  console.log('centre:', scene.bounds.center); // e.g. to frame a camera
+}
+```
+
+`extractThumbnail()` prefers SketchUp's clean `model_thumbnail` over
+`preview_thumbnail`, which has the red/green/blue axis lines drawn in and
+reads as clutter on a catalogue card; `thumb.source` says which was used.
+
+It returns `null` rather than throwing when there is no usable preview.
+That includes **legacy (pre-2021 MFC) files**: those embed PNGs too, but
+the container stores them without entry names, so a thumbnail cannot be
+distinguished from a material's texture image without guessing.
+
+`scene.bounds` is `null` for a model with no geometry, so an empty model
+stays distinguishable from one sitting at the origin.
+`buildInstancedScene()` exposes the same field, computed from the placed
+node transforms so both builders agree.
+
 ### Exporting an instanced GLB
 
 ```typescript
--- a/packages/typescript/src/index.ts
+++ b/packages/typescript/src/index.ts
@@ -31,6 +31,7 @@ import {
 import { isLegacy, parseLegacyToRaw } from './legacy';
 import { encodeIndices } from './gltf-indices';
 import { buildInstancedSceneFromParsed, InstancedScene } from './instanced';
+import { extractThumbnail, SkpThumbnail } from './thumbnail';
 
 export * from './model';
 export type {
@@ -40,6 +41,8 @@ export type {
   LocalPrimitive,
 } from './instanced';
 export { toInstancedGLB } from './instanced-glb';
+export { extractThumbnail } from './thumbnail';
+export type { SkpThumbnail } from './thumbnail';
 export type { InstancedGlbOptions } from './instanced-glb';
 export * from './errors';
 export * from './observability';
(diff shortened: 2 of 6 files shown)


iamahsanmehmood (approved):
Reviewed both features in depth, not just the description.

Thumbnails: traced contents.materialFiles' population back to vff.ts myself (the .endsWith('.png') branch, not a materials-specific filter despite the field name) to confirm meta/model_thumbnail.png and meta/preview_thumbnail.png really do land there - and confirmed the deliberate design specifically avoids the trap it names (a fuzzy "contains thumbnail" search would false-positive on Untitled.skp's real materials/_/thumbnail.jpg texture; looking up the two known entry names exactly does not). Checked the PNG IHDR width/height offsets (16/20, big-endian) against the actual PNG spec layout - correct. The legacy null-return is honest, not a shortcut: legacy's embedded DIBs genuinely have no entry names to distinguish a thumbnail from a texture without guessing, which is a real, disclosed limitation rather than an oversight.

Bounds: verified the baked path's read-back-from-Float32Array approach (matches what a consumer sweeping `positions` itself would get, not the more "accurate" but inconsistent float64 intermediate - good catch, and the 1.4e-6m discrepancy the test caught proves this wasn't theoretical). Verified the instanced path's 8-corner transform is mathematically correct for an affine transform (maps box corners to the transformed box's corners) and confirmed the per-resource local-box cache means shared geometry gets swept once regardless of placement count, matching this feature's own "computed during the bake, costs nothing extra" claim. Matrix convention is consistent with what I already checked in flatten-instanced.ts for #200.

CI green, new fields only (bounds/thumbnail are additive), no breaking changes. Approving.

PR #208, opened 2026-08-22
perf(typescript): store vertex coordinates in a flat Float64Array (1.6x, benchmarked both mappings)

thomazyujibaba:
Implements item 1 of #201, on the terms you set: both mapping strategies implemented behind one interface, benchmarked against the real hot path, and the winner shipped.

## Summary

`GeometryBuilder.vertices` moves from `Map<number, [x, y, z]>` to a flat `Float64Array` plus an id→index mapping.

Every `[x, y, z]` was a separate JS array with its own object header, and every Map entry added hash-bucket overhead — so a million vertices cost **~92 MB to carry 24 MB of coordinates**. That overhead is the dominant term in this package's memory use on large files, and the reason it needs 8–16 GB of heap where the other ports need far less.

`f64`, not `f32` — the file stores doubles and narrowing at parse time would silently change parsed geometry. Holding that as the hard requirement you called it.

## The measurement

Ids are sparse, as you predicted: ~33% dense on aggregate across the fixtures, as low as **7.6%** within a single definition. Raw `Float64Array[id]` would waste more than it saves, so the id→index mapping is mandatory.

Both candidates, at 1M vertices and the ~33% density real files show (RSS after forced GC, values retained, medians of 5):

| storage | memory | vs today | lookup |
|---|---:|---:|---:|
| today: `Map<id,[x,y,z]>` | 92 MB | 1.0x | 1068 ns |
| **`Map<id,index>`** | **57 MB** | **1.6x** | **1088 ns** |
| sorted `Uint32Array` + binary search | 42 MB | 2.2x | 1445 ns |

**`MapVertexStore` ships.** The sorted array saves more memory but costs **~35% on lookup**, and lookup is the hot path — `face-groups.ts` calls `get()` once per triangle corner, so gets vastly outnumber writes. Taking the larger share of the memory win at parity on speed is the better trade for a library whose scene building is already the expensive half.

You said to weight toward maintainability if they were close. They aren't close on lookup, so the measurement decided it — and the simpler implementation happens to be the one that won.

`SortedVertexStore` is **kept, tested and benchmarked** rather than deleted: it wins if lookups ever stop dominating, and keeping it makes the decision re-checkable instead of folklore. `npm run bench:vertex-store` reruns the comparison.

### Honest scope note

This is **1.6x on vertex storage**, not 1.6x on a file's total footprint. The repository's fixtures are too small to show it end-to-end — at 11k vertices the residual `Vertex[]` objects that `buildDefinition()` materialises dominate, and that was item 3, which we agreed to close. So the number above is from the microbenchmark at a realistic scale and density, not from a large real file I don't have. If you have a large `.skp` to hand, I'd genuinely like the before/after RSS on it before this merges.

## Two things worth flagging in review

**`triangulateFace3D` branched on `vertices3D instanceof Map`.** A `VertexStore` is not a `Map`, so it would have silently fallen through to the `Record` path and returned `undefined` for every vertex — no error, just empty geometry. It now checks for a callable `get`, which keeps both existing call shapes working. This is exactly the dual-mode you spotted in your first reply; it needed one more case.

**`get()` returns a fresh array per call, deliberately.** Returning a reused buffer would be the obvious optimization and would be a bug: `face-groups.ts` *retains* the returned array (`group.localVerts.push(p)`), so a shared buffer would corrupt geometry. Left as-is with a comment.

## Testing

`npm test` → **322 passed, 36 files** (303 pre-existing, all unchanged + 19 new).

The existing fixture-parity suite is the real guard here — both scene builders read `builder.vertices` through `face-groups.ts`, and #200's flatten-and-compare test covers both paths on real files. It passed unchanged.

`tests/vertex-store.test.ts` runs the same suite against **both** implementations: sparse ids, ids arriving **out of order** (the sorted store inserts in place — the easiest thing to get subtly wrong), growth past initial capacity, overwrite-without-double-counting, f64 precision preservation, insertion-order iteration (which `buildDefinition()` and the JSON export depend on), and a deterministic 3,000-write randomised pattern asserted key-by-key against a reference `Map`. Plus a test that the two implementations agree with each other on an identical write sequence.

```
$ npm run typecheck        # clean
$ npm run build            # ESM/CJS/DTS success
$ npm test                 # 322 passed (36 files)
$ npm run lint             # clean
$ npm run bench:vertex-store   # table above
```

## Checklist
- [x] Tests pass locally (322/322)
- [x] Documentation updated (module-level rationale in `vertex-store.ts`; no user-facing API change to document)

## Notes

- **No public API change.** `GeometryBuilder` is internal and never re-exported, as you confirmed earlier; `Definition.vertices` is still the same `Vertex[]`, built the same way. No new dependencies.
- Benchmark added as `npm run bench:vertex-store`, following the existing `bench:instanced` convention — needs `--expose-gc` for stable RSS, which the script sets.
- Scope is `packages/typescript`; the other four ports are untouched.
- With this, #201 is: item 2 merged (#202), item 3 closed as not worth pursuing, item 4 no action, item 1 here.

### DIFF
--- a/packages/typescript/package.json
+++ b/packages/typescript/package.json
@@ -20,6 +20,7 @@
     "copy-dist": "node copy-dist.js",
     "test": "vitest run",
     "bench:instanced": "vite-node tests/bench/instanced-bench.ts",
+    "bench:vertex-store": "node --expose-gc ./node_modules/vite-node/vite-node.mjs tests/bench/vertex-store-bench.ts",
     "lint": "eslint src/",
     "typecheck": "tsc --noEmit"
   },
--- a/packages/typescript/src/geometry.ts
+++ b/packages/typescript/src/geometry.ts
@@ -1,6 +1,7 @@
 import { TlvNode, readF64, readU32, parseVarInt, parseTlvRecursive } from './parser';
 import { ParseOptions, emitLog } from './observability';
 import { EdgeFlagStore } from './edge-flags';
+import { DefaultVertexStore, type VertexStore } from './vertex-store';
 
 export interface GeometryBuilderInstance {
   offset: number;
@@ -34,7 +35,9 @@ export interface GeometryBuilderFace {
 }
 
 export class GeometryBuilder {
-  vertices = new Map<number, [number, number, number]>(); // id -> [x, y, z]
+  /** id -> [x, y, z]. Backed by a flat Float64Array rather than a Map of
+   * boxed arrays: see vertex-store.ts for the memory rationale. */
+  vertices: VertexStore = new DefaultVertexStore();
   edges = new Map<number, [number | null, number | null]>(); // id -> [v1, v2]
   /** Edge id -> display flag byte (D307). Backed by a Uint8Array rather
    * than a Map: the value is one byte, and a Map entry costs ~30. */
--- a/packages/typescript/src/legacy.ts
+++ b/packages/typescript/src/legacy.ts
@@ -17,6 +17,7 @@
 
 import { GeometryBuilderFace, GeometryBuilderInstance, ParsedDefinition } from './geometry';
 import { EdgeFlagStore } from './edge-flags';
+import { DefaultVertexStore, type VertexStore } from './vertex-store';
 import { Material, Texture, ParsedRawData } from './model';
 import { SkpParseError } from './errors';
 import { ParseOptions, PROGRESS_INTERVAL, emitLog, emitProgress } from './observability';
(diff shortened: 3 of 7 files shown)


iamahsanmehmood (approved):
Reviewed the implementation directly, not just the benchmark table.

Traced SortedVertexStore's in-place insertion by hand with a concrete example (inserting id=20 into sortedIds=[10,30,50]) to confirm the search()-then-copyWithin-shift-then-write sequence actually preserves sort order rather than just looking plausible - it does. Confirmed copyWithin's overlapping-region behavior is spec-guaranteed safe (memmove semantics), so the self-overlapping shift within one array isn't a hazard. Also confirmed count/sortedCount stay in lockstep by construction (remember() is only ever called once per new id, from within the same set() call that increments count), so the two independently-grown arrays (idList/coords vs sortedIds/slots) can't desync in capacity.

Caught and independently verified the triangulateFace3D fix is the right one: checking for a callable get() instead of instanceof Map is exactly what's needed so a VertexStore doesn't silently fall through to the Record branch and return undefined for every vertex - that would have been a silent empty-geometry failure, not a crash, so good catch finding it yourself before it shipped.

On the decision itself: benchmarking both implementations rather than picking from the sorted-array's theoretical cache-locality advantage was the right call - 1.6x at parity on lookup speed beats 2.2x at a 35% lookup cost for a library whose hot path is exactly that lookup. Keeping SortedVertexStore tested and benchmarked rather than deleting it is good practice - makes the decision re-checkable rather than folklore, per your own framing.

Honest scope note about 1.6x being vertex-storage-only (not whole-file, since item 3's closure means Vertex[] materialization still dominates at small scale) is appreciated - no overclaiming.

CI green, no public API change. Approving.

PR #223, opened 2026-08-25
fix(cpp): normalize coedge orientations

thomasloockx:
## Summary
Normalize coedge orientations produced by the modern and legacy C++ parsers to the documented `+1`/`-1` representation.

The C++ `CoEdge` API documents:

- `+1`: traverse the referenced edge from `v1` to `v2`
- `-1`: traverse it from `v2` to `v1`

The parsers were instead exposing the raw SketchUp sense values, where `0` means forward and `1` means reversed. This violated the public model contract and caused consumers following that contract to interpret otherwise valid face loops as disconnected or incorrectly ordered.

## Type
- [x] Bug fix
- [ ] New feature
- [ ] Documentation

## Changes
- Normalize modern TLV coedge values from raw `0`/`1` to public `+1`/`-1` values while parsing.
- Apply the equivalent normalization to legacy `sense` values.
- Add assertions that every parsed orientation is normalized and every directed edge ends where the next directed edge begins.
- Add [`coedge_orientation_regression.skp`](https://github.com/thomasloockx/openskp/blob/fix/cpp-coedge-orientation-main/packages/cpp/tests/fixtures/coedge_orientation_regression.skp), the model that exposed the issue, as an end-to-end regression fixture. Its 43,833 face loops are all checked.

For example, a loop whose raw SketchUp senses are `0, 1, 0` must be exposed as `+1, -1, +1`. Returning `0, 1, 0` leaves `0` outside the documented API domain and misidentifies the reversed edge.

## Testing
- `cmake --build build/cpp-release -j 4`
- `ctest --test-dir build/cpp-release --output-on-failure` — 144/144 passed
- `cmake --build build/cpp-release --target openskp-format-check`
- `git diff --check`

## Checklist
- [x] Code follows the project style guidelines
- [x] Tests pass locally
- [x] Documentation updated (not applicable; implementation now matches existing documentation)


iamahsanmehmood (approved):
Verified against the documented contract (`CoEdge::orientation` in `model.hpp`: "+1 = same direction, -1 = reversed") — the previous code exposed raw SketchUp sense bits (0/1) on both the modern TLV path and the legacy MFC path, violating that contract. Confirmed the fix's polarity matches: raw A20F 0 → +1 forward (also matches the Python port's already-correct modern-path normalization exactly), legacy sense truthy → -1 reversed.

The new `ExpectConnectedNormalizedLoops` check is a real strengthening of the test suite — it verifies not just that orientation is in {+1, -1} but that consecutive coedges in a loop actually connect (v2 of one matches v1 of the next, accounting for orientation), which is a stronger correctness guarantee than existing tests had. Good catch with the regression fixture.

CI green across all required checks. Merging.

PR #266, opened 2026-09-05
fix(writer): SketchUp can save our files (persistent IDs in one sequence); front_uv pins in SketchUp's basis and scaled by applied size; reader vertical tolerance; face-me flag

tuxiasumari:
Four writer/reader fixes found while chasing "textures look wrong in SketchUp" and "SketchUp can't save the file" on real IngeTrazo exports, each measured against the SketchUp SDK (`SUMeshHelperGetFrontSTQCoords` / `SUModelSaveToFile`) before and after. The persistent-ID one (item 5 below) is the one I'd merge first: every non-trivial file this writer produces opens in SketchUp but cannot be saved there.

## What was wrong

Two defects in the Python writer's texture positioning. Neither shows in the
existing tests because every test square is axis-aligned with its first edge
along `Z × n`, and every test material is 1 inch per tile.

1. **Basis.** `_uv_matrix_for_face` solved the 3×3 in a basis built from the
   face's *first edge* (`points[1] - points[0]`, normal × that). Real SketchUp
   (and openskp's own reader, `_face_groups.face_uv_basis`, calibrated against
   SDK-authored files) express and invert the matrix in a basis derived from
   the **normal alone**: `U = normalize(Z × n)`, `W = n × U` (`(X, ±Y)` for a
   horizontal face). The two agree only when the first edge happens to run
   along `Z × n` — otherwise the texture comes out turned by the angle between
   them. A horizontal square listed from its second corner arrives in SketchUp
   rotated 90°; from its third, 180°; a curved surface of many small quads
   shatters, each quad turned by its own first edge.

2. **Scale.** SketchUp stores the matrix in *inches of texture space* and
   divides by the material's applied width/height when it reads a face's UV
   back (`compute_face_uv` already does exactly that). The caller's pins are
   in tiles and were fitted as-is, so a material applied at 2 m per tile
   (78.74 in) came out 78.74× too big on every positioned face: a pool's
   water as one flat blue slab. `openskp.edit` and `to_python_code` were
   already dodging this by forcing `applied_height=1.0`, which throws the
   material's real size away.

## How it was measured

Exported 11 control squares (horizontal with the first edge along +X, +Y, −X
and at 30°; walls facing ±X and ±Y, one of them from a vertical and from a
horizontal first edge; a 45° slope), converted the `.skp` with the SDK
(`SUMeshHelperGetFrontSTQCoords`, via a small skp2dae tool) and fitted the
SDK's UVs against the pinned ones per face:

| pins as written by 1.2.0 | scale | turn |
|---|---|---|
| first edge +X (the test case) | 0.100 | 0° |
| first edge +Y | 0.100 | 90° |
| first edge −X | 0.100 | 180° |
| first edge at 30° | 0.100 | 30° |
| +Y wall, first edge vertical | 0.100 | 90° |
| +Y wall, first edge horizontal | 0.100 | 180° |

(0.100 = 1 / the 10-inch applied size.) With this branch: scale 1.000, turn
0° on all eleven.

3. **Vertical tolerance and the downward basis** (reader and, through it, the writer).
   `face_uv_basis` snapped to the world axes only within 1e-9 of vertical and gave
   a face looking down `(X, −Y)`. Measured with the SDK on faces tilted from 1e-10
   to 1e-2, looking up and down: SketchUp keeps the world axes while the sine of
   the tilt is below 1e-3 (1.0000004e-3 still vertical, 1.0001e-3 already the
   cross product), and a downward face gets `(−X, +Y)`, the 180° turn of the
   upward basis. A horizontal face whose stored normal carries float noise, and
   every underside of a real model, read back turned.

4. **Face-me flag (writer).** Unrelated to UVs but tiny and useful: the reader
   already decodes SketchUp's behavior byte (`always_faces_camera`,
   `shadows_face_sun`); the writer left it zero. `add_component_definition`
   now takes both flags, `edit` replays them.

5. **Files could not be saved by SketchUp** (the one I'd merge first). Every section's
   `_ArchiveWriter` started pids at 1, and `to_bytes` patched the header's pid counter
   (a u32, treated as u16) by materials + layers only. SketchUp loads the file,
   renumbers the duplicate pids, and then `SUModelSaveToFile` returns
   `SU_ERROR_SERIALIZATION` (7) — "Save failed" in SketchUp Web on every export
   larger than a toy. Minimal repro: a 1-face definition then a 3-face one, both
   placed. Bisected with the SDK: patching the counter alone rescues small files
   and fails past 65 535 pids or when a definition comes first; one sequence
   across sections fixes all of it (7 MB and 27 MB real exports re-save).

## The fix

- `_uv_matrix_for_face` solves in `face_uv_basis(normal)`; `_face_uv_basis`
  (first-edge) is gone.
- `SkpBuilder` records each textured material's applied size at
  `add_texture_material` time; both `add_face`s scale `front_uv`/`back_uv`
  by it (`_scale_pins`). Pins stay in tiles for the caller, as documented.
- `openskp.edit._replay_materials` and `codegen` write the source's real
  applied size again.
- Pids: `SkpBuilder._pid_start` from the scaffold counter, each `_ArchiveWriter`
  continues from `_next_pid()`, `to_bytes` writes the last pid as a u32.
- `_face_groups.face_uv_basis`: `VERTICAL_TOLERANCE = 1e-3` and `(−X, +Y)` for a
  downward normal (the writer inherits both through `_uv_matrix_for_face`).
- Docstrings of `_solve_uv_matrix`, `add_texture_material`, `add_face`,
  `Face.uv_transform` updated; CHANGELOG entry under Unreleased.

## Tests

- `test_vertex_order_does_not_turn_the_mapping`: the same square listed from
  another corner stores the same `diag(50, 50)`.
- `test_pins_read_back_as_pinned_on_every_orientation`: on six orientations
  and vertex orders, pinned to a deliberately skewed map, the reader hands
  back exactly the pinned UV at each pin and the affine value at the fourth
  corner.
- `test_pins_are_in_tiles_whatever_the_applied_size`: applied size 10,
  `(50,0,0) → (1,0)` stores `diag(5, 5)` and reads back as `u = 1`.
- `test_persistent_ids_run_in_one_sequence_across_sections`, and the real-SketchUp
  `test_a_small_definition_before_a_bigger_one_saves_through_real_sketchup`
  (`SUModelSaveToFile` on the minimal repro).
- `test_face_uv_basis_snaps_and_turns_like_real_sketchup`: the measured tolerance
  and the downward basis.
- `TestRealSketchUpOracle::test_positioned_texture_survives_vertex_order_through_real_sketchup`:
  same rotated vertex order through `SUUVHelperGetFrontUVQ` (needs the SDK
  DLL, skipped otherwise).
- The existing ground-truth tests (`diag(50,50)`, the tilted asymmetric
  byte-exact one) still pass: their faces all have the first edge along
  `Z × n`, where both bases coincide — which is why the defect hid.

Full suite: all green locally (see the run in the message).

https://claude.ai/code/session_013JHz7hRbUdcRFiDS8Z1LYv

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,13 @@ test square happened to be axis-aligned and every test material 1 inch per tile:
   `openskp.edit` and `to_python_code` write the source's real applied size again instead of forcing
   `applied_height=1.0` to dodge the double division.
 
+- **Vertical tolerance and the downward basis (reader and writer).** `face_uv_basis` treated a
+  normal as vertical only within 1e-9 of it, and gave a face looking down the basis `(X, −Y)`. Real
+  SketchUp, measured with the SDK on faces tilted from 1e-10 to 1e-2 looking up and down, keeps the
+  world axes while the sine of the tilt is below 1e-3, and turns the basis 180° for a downward face:
+  `(−X, +Y)`. A horizontal face whose stored normal carries float noise, and every underside of a
+  model, read back (and were written) turned. `VERTICAL_TOLERANCE = 1e-3` in `_face_groups`.
+
 Measured through the SDK's own `SUMeshHelperGetFrontSTQCoords` (skp2dae) on 11 orientations before and
 after; new tests pin the invariant — what you pin is what the reader hands back at that point, on
 six orientations and vertex orders, and at applied size 10 — plus a real-SketchUp oracle test for the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed — files the writer produced could not be SAVED by SketchUp (Python writer)
+
+Every section's writer (materials, layers, definitions, root geometry) handed out persistent IDs
+from 1 again, so a material, a definition and a root instance could all carry pid 1, and the header's
+pid counter — a 32-bit field the writer treated as 16-bit — only grew by the material and layer count.
+SketchUp loads such a file, renumbers the duplicates, and then `SUModelSaveToFile` fails with
+`SU_ERROR_SERIALIZATION` once the model is big enough or a definition happens to come first: a real
+7 MB export opened fine in SketchUp Web and every attempt to save it ended in "Save failed". Measured
+with the SDK: a 1-face definition followed by a 3-face one is enough. Pids now run in one sequence
+across sections, continuing from the scaffold's counter, and the counter is written as the u32 it is,
+with the last pid handed out. Real exports of 7 MB and 27 MB re-save through the SDK.
+
 ### Added — face-me components in the writer (Python)
 
 `add_component_definition(name, always_faces_camera=True, shadows_face_sun=True)` sets SketchUp's
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added — face-me components in the writer (Python)
+
+`add_component_definition(name, always_faces_camera=True, shadows_face_sun=True)` sets SketchUp's
+component-behavior byte the reader already decoded (bit 0 / bit 1 of byte −9 in the definition's
+43-byte gap), so a 2D person or cut-out tree written with openskp turns toward the camera in
+SketchUp instead of standing still. `openskp.edit` replays a source file's flags when re-saving.
+
 ### Fixed — `front_uv`/`back_uv` pins land where SketchUp draws them, on any face and at any applied size (Python writer)
 
 Two defects in the writer's texture positioning, invisible to its own reader tests because every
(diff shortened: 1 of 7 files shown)


iamahsanmehmood (approved):
Reviewed the full diff, not just the description.

The persistent-ID fix (the one to merge first, per the PR description) is correct: `_next_pid()` correctly walks writers in reverse-creation order to find the true next-free pid regardless of which sections exist, `to_bytes()` now writes the counter as the real u32 it is instead of the old truncated u16 write, and the fix comes with a real `SUModelSaveToFile` oracle test on the exact minimal repro described (a 1-face definition before a 3-face one) - this isn't a guess, it's SDK-verified the same way this project requires.

The UV basis/scale fixes are well-reasoned and quantitatively measured (11 orientations against `SUMeshHelperGetFrontSTQCoords`, before/after tables in the description) - the root cause (first-edge-derived basis vs. SketchUp's actual normal-derived basis) is a believable, specific explanation for why every existing test passed despite the bug (all test faces happened to have their first edge along `Z × n`). New tests correctly target the axis-aligned blind spot directly (vertex-order invariance, six orientations, applied-size scaling).

Face-me flag addition is small, matches the reader's already-decoded bits exactly.

CI green across all 5 languages (only Python changed; others correctly just ran their own unaffected suites).

Python-only, as usual for a single-language contributor PR - cross-language porting to track separately, not block this merge.

PR #276, opened 2026-09-08
feat(ts): add toFragments() exporter for Fragments (.frag) binary 3D/BIM streaming

nooraliqureshi:
### Summary

Adds direct export support for ThatOpen Engine's Fragments (`.frag`) binary format to the TypeScript package (`toFragments` / `SkpFile.prototype.toFragments`).

Fragments is an instanced, FlatBuffers-based binary container (deflated by default) designed for streaming and GPU-instanced 3D/BIM rendering in `@thatopen/fragments` and IFC.js viewers.

### Implementation Details

- **Official FlatBuffers Schema Bindings**: Uses TypeScript bindings generated directly from ThatOpen's canonical `index.fbs` schema (`ThatOpen/engine_fragment`) in `packages/typescript/src/_fragments_fb/`, replacing hand-rolled schemas.
- **Instancing & Deduplication**: Preserves SketchUp component instancing so placements sharing the same definition share one `Shell` and `Representation`, keeping payload sizes minimal.
- **TRS Decomposition & Scale/Mirror Baking**:
  - Implements `decomposeTrs` to extract translation, unit direction vectors (`xDir`, `yDir`), scale magnitudes, and negative-determinant mirroring (`det < 0`).
  - Implements `bakePrimitive` and `scaleCacheKey`: `Transform` direction vectors remain unit length, while non-uniform scaling and mirroring are baked into local primitive vertices.
  - Mirrored instances invert triangle winding (`[a, c, b]`) to preserve correct face normals.
  - Instances sharing the same `(definition, scale, mirror)` cache key deduplicate onto a single shared `Shell`.
- **Spatial Structure & Metadata**:
  - Encodes the scene graph hierarchy into `SpatialStructure` nodes with `localId` references.
  - Exports display names via `Attribute` data triples (`["Name", value, "STRING"]`) and preserves per-layer visibility state in `Model.metadata`.
- **Deflate Compression**: Matches ThatOpen's default RFC 1950 zlib-compressed wire format (`fflate.zlibSync`), with opt-in uncompressed FlatBuffers via `{ raw: true }`.

### Public API

```typescript
import { SkpFile, buildInstancedScene, toFragments } from 'openskp';

// Option 1: Direct file wrapper
const skpFile = SkpFile.open('model.skp');
const fragBytes = skpFile.toFragments();

// Option 2: From an InstancedScene
const scene = buildInstancedScene(skpArrayBuffer);
const fragBytes = toFragments(scene, {
  raw: false,         // default: false (zlib deflated)
  doubleSided: false, // default: false
});
```

### Changes

- `packages/typescript/src/_fragments_fb/`: Official generated FlatBuffers TypeScript bindings from `index.fbs`.
- `packages/typescript/src/fragments.ts`: Exporter implementation with TRS matrix decomposition and scale/mirror geometry baking.
- `packages/typescript/src/index.ts`: Export `toFragments`, `FragmentExportOptions`, and `SkpFile.prototype.toFragments`.
- `packages/typescript/tests/fragments.test.ts`: Test suite covering FlatBuffers runtime parsing (`Model.getRootAsModel`), identifier check (`0001`), shared geometry deduplication, unit direction vectors, scale/mirror baking, and real fixtures (`SU_File.skp`, `gondola_v20.skp`, `Untitled.skp`) with no silent skips.

### Verification

- `npx vitest run tests/fragments.test.ts`: 20/20 tests passing.
- `npm run typecheck`: 0 type errors.
- `npm run lint`: 0 errors, 0 warnings.
- `npm run build`: Bundles clean CJS, ESM, and DTS declarations.
- Verified against all CI suites (TypeScript, Python, C++, Dart, .NET).


iamahsanmehmood (changes requested):
Thanks for building this - Fragments support for the TypeScript package is genuinely valuable. This PR predates the Python side's own Fragments work (`packages/python/src/openskp/export/fragments.py`, merged after this PR was opened), which went through a fairly thorough correctness/verification process against the real ThatOpen ecosystem and found some non-obvious gotchas along the way. Pointing those out here so this PR can land at the same bar rather than duplicate the same lessons the hard way.

## 1. Hand-rolled FlatBuffers schema, not verified against ThatOpen's real one

`fragments-schema.ts` hand-writes the FlatBuffers builder calls (field slots, vector types, etc.) by hand instead of using generated bindings. That's very easy to get subtly wrong - a single wrong vtable slot index or field type produces a `.frag` file that looks valid (has a plausible root offset, inflates fine) but is silently corrupt or unreadable by the real `@thatopen/fragments` library. The Python side avoids this by vendoring bindings actually generated by `flatc` from ThatOpen's own published schema (`packages/python/src/openskp/_fragments_fb/`, generated from `ThatOpen/engine_fragment`'s `packages/fragments/flatbuffers/index.fbs`). Worth doing the same here - generate real TS/JS bindings from that same `.fbs` file (or use `@thatopen/fragments`' own exported serialization internals, if it exposes any) rather than a hand-transcribed schema.

## 2. Never verified against the real `@thatopen/fragments` runtime

All 5 current tests check that the output is non-empty, inflates without throwing, and that reading the first 4 bytes as an int32 gives a plausible-looking root offset - none of them actually parse the result back, not even with this PR's own schema. That doesn't catch wrong field encoding, wrong material colors, wrong geometry, wrong instancing, etc. - it only proves "produces *a* flatbuffer," not "produces *the right* flatbuffer." The Python module's real bar (documented in its own module docstring) is: load the output through the actual, unmodified `@thatopen/fragments` npm package (`SingleThreadedFragmentsModel`) and confirm it triangulates/renders correctly, including real instance dedup and rotation - that's what actually proves interop, and it's how a real bug (see #6 below) got caught before merge instead of after. Given this is TypeScript, that verification is even more directly reachable here than it was from Python - `@thatopen/fragments` is a JS/TS package.

## 3. No scale/mirror handling - Transform has no scale field

`fragments.ts` takes `xDirection`/`yDirection` straight from the world matrix's columns (`m[0..2]`, `m[4..6]`) with no normalization or scale extraction. Fragments' `Transform` struct has no scale field at all (confirmed by reading the real importer's own serialization code) - so any instance with non-uniform scale, or a mirrored (negative-determinant) transform, will come out wrong: direction-vector magnitude alone can't represent that, and a rotated non-uniform scale corrupts the shape outright if the geometry itself isn't also rescaled. Python's fix (`_decompose_trs`/`_bake_primitive` in `fragments.py`) bakes each instance's own scale/mirror into its own geometry variant before writing, so instances sharing the same (definition, scale) still dedupe and share one baked shell, and only genuinely distinct scale factors get their own copy. Worth the same approach here rather than passing scale through silently.

## 4. Tests silently no-op when fixtures are missing

```ts
if (fs.existsSync(fixturePath)) { ... }
```
in 3 of the 5 tests means the test body - and every assertion in it - simply doesn't run if the fixture isn't present, but the test still reports "passed." If those fixtures aren't actually committed under `packages/typescript/tests/fixtures/`, these tests may be doing nothing at all in CI. Should fail loudly (`expect(fs.existsSync(fixturePath)).toBe(true)` or just don't guard it) if a fixture a test depends on isn't there, not silently skip.

## 5. Read direction missing (`fromFragments`)

The Python module is bidirectional now - `to_fragments`/`export` (write) and `from_fragments`/`read` (read any real `.frag` file, not just this project's own output). For real cross-language parity eventually, TypeScript would need the same. Not blocking this PR, but worth tracking as a follow-up once the writer here is solid.

## 6. A concrete gotcha to save you time if/when you do add reading

When we added the Python reader, everything passed against synthetic fixtures and even ThatOpen's own small sample file - and then broke on a real file converted through ThatOpen's actual `IfcImporter`, with an `IndexError`. Root cause: `Representation`'s own position in the `Representations` vector is **not** the same as `Representation.Id()` - `Id()` is the real index into `Meshes.Shells`, and only OpenSKP's own writer happens to keep the two equal. The real reader resolves shells via `meshes.shells(repr.id!, ...)` (`@thatopen/fragments`' own `Utils/edit/fetch-functions.ts`), not by vector position. Mentioning this now so it doesn't cost you the same debugging loop later.

---

Happy to share more detail on any of the above, or take a look at an updated version. `packages/python/src/openskp/export/fragments.py` and `packages/python/tests/test_fragments.py` are the closest thing to a reference for what "done" looks like for this format in this repo - both the writer's own verification claims and the reader's are stated plainly there, gaps included.

nooraliqureshi:
Thanks for the detailed review and pointers, @iamahsanmehmood! I've updated the PR to address all items:

1. **Official FlatBuffers Bindings**: Removed the hand-rolled schema and generated TypeScript bindings directly from `index.fbs` into `packages/typescript/src/_fragments_fb/`. All serialization now uses the official generated classes and structs.

2. **TRS Decomposition & Scale/Mirror Baking**:
   - Implemented `decomposeTrs` to separate translation, unit-length direction vectors (`xDir`, `yDir`), scale magnitudes, and negative-determinant mirroring.
   - Implemented `bakePrimitive` and `scaleCacheKey` matching Python/C++:
     - `Transform` direction vectors remain normalized unit vectors.
     - Scales and mirroring are baked into local vertex coordinates.
     - Triangle winding is reversed (`[a, c, b]`) on mirrored instances to prevent inside-out inverted normals.
     - Placements sharing the same `(definition, scale, mirror)` state deduplicate onto a single `Shell`, while distinct scales produce separate shells.

3. **FlatBuffers Runtime Verification & Test Hardening**:
   - Removed the silent `if (fs.existsSync)` guards so missing fixtures fail loudly.
   - Added assertions that parse emitted bytes back through `Model.getRootAsModel`:
     - File identifier check (`0001`).
     - Shell deduplication vs sample count.
     - Transform unit-length vectors and position accuracy.
     - Material color fidelity and bounding box validity.
     - Scale baking, winding flip, and multi-scale shell generation tests.
   - Verified cleanly against `SU_File.skp`, `gondola_v20.skp`, and `Untitled.skp`.

4. **Follow-up Note**:
   - Appreciate the heads-up on `Representation.id` indexing into `Meshes.shells` rather than vector index; will make sure that's accounted for when adding `fromFragments`.

iamahsanmehmood (changes requested):
Re-reviewed after the update — this is a substantial rewrite (37 files, +4,211 lines) and it addresses most of the original findings well. Went through each:

**1. Hand-rolled FlatBuffers schema → fixed.** `packages/typescript/src/_fragments_fb/index.fbs` is now byte-identical to Python's own vendored `index.fbs` (only the `namespace` differs, as expected: `openskp.fragments_fb` vs `openskp._fragments_fb`). Spot-checked a generated binding (`representation.ts`) against Python's equivalent — field offsets match exactly (`id()` at 0, `representationClass()` at 28, `sizeOf()` = 32). This is genuinely the canonical ThatOpen schema now, not something reverse-engineered.

**2. No verification against the real `@thatopen/fragments` package → still open.** I checked `package.json` (`dependencies`: `earcut`, `fflate`, `flatbuffers`; `devDependencies`: eslint/tsup/typescript/vitest tooling only) and all imports in `tests/fragments.test.ts` — everything validates output exclusively through this PR's own generated `_fragments_fb` bindings (`Model.getRootAsModel`, `.shellsLength()`, etc.), never through the actual `@thatopen/fragments` npm package (no `SingleThreadedFragmentsModel` or equivalent anywhere).

That's an important distinction: self-consistency checks prove the writer and the generated reader agree with each other, but not that a real ThatOpen consumer can open the file. This is exactly the gap that caught the `Representation.Id()` indexing bug on the Python port originally — Python's own bindings would have happily read back its own (buggy) output too. Since `@thatopen/fragments` is a real, published npm package, I'd like to see at least one test that round-trips through it directly (e.g. add it as a devDependency and load one of the fixture outputs with `SingleThreadedFragmentsModel` or whatever the current API calls for). That's the strongest evidence of genuine interop, and it's a fairly small addition given how much else here is already solid.

**3. No scale/mirror handling → fixed.** `decomposeTrs` (determinant-based mirror detection), `bakePrimitive` (per-axis scale baking, X-negation + triangle winding flip `[a, c, b]` for mirrored instances), and `scaleCacheKey` (folds mirror into the sign of X for dedup) are all present and match the architecture already proven correct in the Python/C++ ports. Covered by dedicated tests (`TRS Decomposition and Scale/Mirror Baking`), including one that confirms mirrored + non-uniform-scaled instances still dedupe correctly onto shared baked shells.

**4. Tests silently skipping when fixtures are missing → fixed.** The new `Real SketchUp Fixture Tests (No Silent Skips)` block runs unconditionally against `SU_File.skp`, `gondola_v20.skp`, and `Untitled.skp` with no `existsSync` guards, and does real structural assertions (non-zero shells/samples/representations/materials, unit-length transform direction vectors, non-inverted bounding boxes, samples > shells for repeated instancing). This is good, meaningful coverage — it's just validating against the PR's own bindings rather than the real ThatOpen library (see #2).

**5. No reader (`fromFragments`) → confirmed still out of scope**, and that's fine — this PR is export-only by design, matching the Python port's own sequencing (writer first, reader as a separate follow-up).

**6. `Representation.Id()` gotcha** (a representation's position in the `Representations` vector isn't necessarily its `Id()` — real readers resolve shells via `meshes.shells(repr.id!, ...)`, not by vector position) — not yet relevant since there's no reader here, but flagging it now so it's not forgotten: when a `fromFragments` reader is eventually built for TypeScript, make sure shell lookups go through `Id()` rather than array position, since only this project's own writer happens to keep the two in sync.

**Overall**: findings #1, #3, #4 are solidly fixed, #5/#6 are non-issues for an export-only PR. #2 is the one thing I'd still want closed before merge — a single real round-trip through `@thatopen/fragments` would give real confidence this interoperates with the actual ecosystem it's meant for, not just with itself. Happy to approve once that's added.

nooraliqureshi:
Thanks for the clear follow-up, @iamahsanmehmood! Item #2 is now resolved and verified:

1. **`@thatopen/fragments` Added as DevDependency**:
   - Added `@thatopen/fragments` (`^3.4.7`) along with its peer dependencies (`three`, `web-ifc`) to `devDependencies` in `packages/typescript/package.json` and updated `package-lock.json` for reproducible `npm ci`.
   - Runtime `dependencies` remain untouched with zero additional overhead.

2. **Real Ecosystem Round-Trip & Interop Verification**:
   - Added a dedicated test suite in `tests/fragments.test.ts` exercising `SingleThreadedFragmentsModel`:
     - **`SU_File.skp`**: Directly instantiates `new SingleThreadedFragmentsModel(...)` from deflated `.frag` bytes; verifies `getItemsIds()`, `getLocalIds()`, item attribute metadata (`model.getItemsData()`), and geometry decoding through ThatOpen's internal engine via `model.getItemsGeometry()` (Three.js `Matrix4` transforms, Float32Array vertex positions, Uint16Array indices, and Int16Array normals).
     - **Raw Buffer Mode**: Validates uncompressed buffer loading (`{ raw: true }`).
     - **`gondola_v20.skp`**: Tests complex multi-instance assembly through ThatOpen's model reader, asserting multi-chunk geometry extraction and layer category preservation (`_category`).

3. **CI & Hygiene**:
   - Rebased cleanly on latest `upstream/main`.
   - Full test suite in `tests/fragments.test.ts` passes (23/23).
   - `npm run typecheck` and `npm run lint` pass with 0 warnings or errors.

iamahsanmehmood (approved):
Verified independently (checked out the branch into an isolated worktree, `npm ci`, ran the full suite + lint + typecheck fresh, not just relying on CI output):

- `package.json`: `@thatopen/fragments` (^3.4.7) genuinely added as a devDependency with its real peer deps (`three`, `web-ifc`) - runtime `dependencies` untouched.
- The new `describe('Real @thatopen/fragments npm package interop (SingleThreadedFragmentsModel)', ...)` block does exactly what I was asking for: constructs the real `SingleThreadedFragmentsModel` from our generated `.frag` bytes and pulls geometry and metadata back out through **ThatOpen's own decoding engine**, not our own bindings - `getItemsGeometry()` (positions/indices/transform), `getItemsData()` (checks real field values like `Name` and `_category`), both compressed and raw (`{ raw: true }`) modes, across two real fixtures including a complex 81-item instanced model (`gondola_v20.skp`). That's the strongest evidence of genuine interop this PR could realistically carry - if there were a `Representation.Id()`-style indexing bug, this is exactly the kind of test that would catch it, since ThatOpen's own reader doesn't share any assumptions with our writer.
- Local results: 397/397 tests passing (41 files), `npm run lint` clean, `npm run typecheck` clean. CI on the latest commit (`76df77e`) is green across the board.

All the findings from the original review are now addressed:
1. Canonical FlatBuffers schema (not hand-rolled) - confirmed
2. Real ecosystem round-trip through `@thatopen/fragments` - confirmed, this update
3. TRS decomposition + scale/mirror baking - confirmed
4. Real-fixture tests with no silent skips - confirmed
5. No reader (`fromFragments`) - correctly out of scope for an export-only PR
6. `Representation.Id()` indexing gotcha - not applicable without a reader; flagged for whenever one is built

Nice work getting this all the way through - approving.

PR #292, opened 2026-09-09
perf(typescript): keep the writer's archive in a typed array, not a number[]

tdurand:
Hey ! Thanks for the library, very useful ! just to state that I did not write this code. It was written by Claude while working on the `.skp` exporter of a project of mine, which uses openskp in the browser. I have read the change, it makes sense to me, and I ran your typecheck, lint and test suite on it (below). I am submitting it because the fix is small and the measurements are real, but please review it as you would any external contribution, and tell me if you would rather I close it and just file the numbers as an issue.

## Summary

`ArchiveWriter` accumulates the archive in a plain `number[]`, and `SkpBuilder.toBytes()` then builds a second `number[]` of the whole file before `Uint8Array.from` makes a third copy. In V8 a `number[]` of byte values costs about 9 bytes of heap per element (measured: 20M elements = 170 MB in Node 23; roughly half that in Chrome with pointer compression), so the archive is held at 9x + 9x + 1x its own size, plus the doubling growth of each array. For a city-sized model that is the difference between a download and a dead tab.

This PR replaces the buffer with a growable `Uint8Array` and makes `toBytes()` collect its pieces and assemble one exact-size result. Output is byte-identical.

## Type
- [x] Bug fix (memory: a large write exhausts the heap)
- [ ] New feature
- [ ] Documentation

## Changes

- `src/create.ts`
  - `GrowableBytes`: a `Uint8Array` that doubles on demand, with `push(...values)` for the record writers' few-byte writes, `append(buffer)` for whole buffers, and a copy-free `view()`. Kept internal; exposed on `_internal` for tests. The class doc keeps the warning the old `appendAll` carried: spreading a large buffer into a call blows the engine's argument limit, so buffers go through `append`.
  - `ArchiveWriter.bytes` is a `GrowableBytes`; the two `writeU32At` back-patch sites write through `.buf` (its signature now accepts `number[] | Uint8Array`).
  - `toBytes()` pushes its parts as `Uint8Array`s and assembles them into one exact-size buffer. `appendAll` has no callers left and is removed.
- `tests/create.test.ts`: the four "Slot-boundary encoding" tests that compared `w.bytes` to an array literal now read `Array.from(w.bytes.view())`. Same assertions.
- `tests/growable-bytes.test.ts`: push across a capacity doubling, append of typed and plain arrays, `view()` as a live window (what the in-place back-patches rely on), growth past 1 MB.
- `CHANGELOG.md`: entry under `[Unreleased]`.

No public API change: `index.ts` re-exports `create` by name, so `GrowableBytes` is not published.

## Measurements

Real exporter (Babylon.js NullEngine scene → openskp), heap sampled every 20 ms, Node 23:

| scene | .skp size | peak heap, 1.1.0 as published | peak heap, this change |
|---|---|---|---|
| 124k triangles (100k terrain + 2,000 boxes) | 20 MB | +464 MB | +87 MB |
| 396k triangles (300k terrain + 8,000 boxes) | 62 MB | +2,144 MB | +184 MB |

The residual is close to the file size plus the writer's vertex/edge registries. Wall time also dropped (6.2 s → 4.8 s on the larger scene), presumably from less GC.

Output equivalence: the files from the patched writer were parsed back with the unmodified `parseSkp`; face, edge and vertex counts and a checksum over every vertex coordinate match the files from the unpatched writer, on both scenes and on the exporter's fixture suite.

## Testing

In `packages/typescript` on `main` (f86957e):

- `npm run typecheck`: clean
- `npm run lint`: clean
- `npm test`: 373 passed, 0 failed (40 files).

## Checklist
- [x] Code follows the project's style guidelines (matched the surrounding code; no formatter run, since the repo has no prettier config and untouched files do not pass `prettier --check` either)
- [x] Tests pass locally
- [x] Documentation updated (CHANGELOG)


iamahsanmehmood (approved):
Reviewed the diff carefully (GrowableBytes: reserve/push/append/view, the writeU32At widening, every out.push() call site in toBytes() and their operand types, the final exact-size assembly loop) - no correctness issues found. Output equivalence is well-argued (byte-identical per the PR description, and the updated "Slot-boundary encoding" tests exercise the same assertions through `.view()`).

Verified independently on the pushed branch (not just trusting the PR description):
- `npm run typecheck`: clean
- `npm run lint`: clean
- `npm test`: 373/373 passing

Resolved a CHANGELOG.md merge conflict against current main (my own recent C++ work landed in the same `## [Unreleased]` section concurrently) by merging main into this branch and pushing the resolution here - no changes to the substance of this PR, purely the merge + a whitespace fix.

Real, honest measurements, thorough testing methodology, and transparent about AI involvement in the original authorship. Good contribution - thank you!

PR #300, opened 2026-09-10
feat(cpp): add WebAssembly build target and streaming definition parser (#264)

### LINKED ISSUE #264: SystemError: error return without exception set — SkpFile.parse() crash on large (~359MB) file, not a size limit
## Summary

`SkpFile.parse()` (`_core.full_parse()`) crashes with `SystemError: error return without exception set` on a specific large (~359MB) `.skp` file, in production, on `openskp==1.1.0`.

## Environment

- `openskp` version: `1.1.0`
- Python: `3.12.3`
- OS: Ubuntu 24.04 (production server)
- Install: `pip install openskp==1.1.0` (system-wide, no venv)

## What happened

A ~359MB SketchUp file (`376,874,735` bytes) failed during parsing after running for ~86 seconds (not an immediate failure):

```
Traceback (most recent call last):
  File ".../process_model.py", line 1114, in convert_skp_to_glb
    skp.parse()  # populates skp._parsed, reused by every export call below
    ^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/openskp/model.py", line 535, in parse
    parsed = _core.full_parse(str(self.path))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
SystemError: error return without exception set
```

Calling code (simplified):

```python
from openskp import SkpFile

skp = SkpFile.open(input_path)
skp.parse()  # <- crashes here, ~86s in
```

## Why I don't think this is a plain size limit

- No OOM: the host had 10GB+ RAM free throughout (`free -h` checked immediately after the failure, no `dmesg`/`journalctl` OOM-killer entries in the relevant window).
- Your own README/docs state verified support for files up to 620MB with 150,000+ definitions — this file is well under that (359MB).
- It ran for ~86 seconds before failing, suggesting it got well into parsing rather than failing immediately on open.

My best guess (unconfirmed) is something specific to this file's internal structure — maybe one unusually large/complex individual definition, given the docs note peak memory is bounded by the largest single definition rather than the whole file.

## `SystemError: error return without exception set` specifically

This is a CPython-level message that fires when a C-level call returns an error/NULL without setting a corresponding Python exception — usually points to a native/C-extension code path that hit an internal failure without properly reporting it as a Python exception. Given `_core.full_parse` is presumably compiled/native, this looks like it's coming from there rather than anywhere in the pure-Python layer.

## Our pipeline, for context (nothing on our side touches raw SKP data)

In case it's useful to know exactly what's calling into `openskp` and whether anything on our end could be pre-processing/mangling the input before it reaches you — it isn't. Our pipeline is:

1. `SkpFile.open(input_path)` — openskp opens the raw file directly, untouched.
2. `skp.parse()` — openskp's own native parser reads the entire file. **This is exactly where the crash happens**, before any of our code runs.
3. `openskp.export.glb.export(skp, output_glb)` — openskp's own exporter, converts the parsed data to GLB + metadata JSON + thumbnail.
4. `openskp.export.ifc.export(...)` — openskp's own exporter again, this time to IFC (we feed this into a separate downstream tool for a streaming format).

Everything on our side happens *after* these calls, operating only on files openskp has already produced (reshaping the metadata JSON into our own schema, running `gltf-transform` on the GLB, etc.) — we have no custom code that reads raw `.skp` binary data at all. So there's no pre-processing step on our end that could be producing a malformed input; the file reaching `skp.parse()` is exactly the original upload, byte-for-byte.

## Impact

This crashes the whole parse with no partial result and no actionable error from the library's side (the `SystemError` itself isn't useful to an end user) — currently handling this on our side by catching it and showing a generic "try re-exporting with Purge unused" message, but obviously a real fix (or at least a clean, catchable exception instead of `SystemError`) on your end would be much better.

Happy to provide more detail, run additional diagnostics against this file, or test a fix if useful.

nooraliqureshi:
## Summary
This PR adds WebAssembly support and brings the streaming definition parser to OpenSKP's C++ package:

1. **WebAssembly (WASM) Target & Emscripten API (`feat(cpp)`):**
   - Adds `OPENSKP_BUILD_WASM` CMake option compiling `libOpenSkp` to WebAssembly (`openskp.wasm` ~830 KB).
   - Exposes zero-overhead Emscripten bindings in `packages/cpp/src/wasm_api.cpp`:
     - `parseSkpToFragments(buffer)`: exports directly to ThatOpen Fragments via `openskp::to_fragments(scene, /*raw=*/true)`.
     - `parseSkpToGLB(buffer)`: exports to GLB via `openskp::to_instanced_glb`.
   - Browser & MV3 CSP compliance: built with `-sDYNAMIC_EXECUTION=0`, `MODULARIZE=1`, `EXPORT_ES6=1`, and dynamic memory growth.
   - Includes cross-platform build scripts `build-wasm.bat` and `build-wasm.sh`.

2. **Streaming Definition Parser for Large Models (`perf(cpp)`) - Resolves #264:**
   - Brings C++ memory parity with the Python streaming parser for large SketchUp models:
     - Direct ZIP memory extraction: replaces `mz_zip_reader_extract_to_heap` with direct `mz_zip_reader_extract_to_mem`, eliminating duplicate heap allocations.
     - Streaming `F901` definition unwrap: mirrors Python's `_unwrap_definitions_container` and `iter_top_level_lazy` to unwrap component definitions upfront, bounding peak memory to one definition at a time and ensuring monotonic progress reporting.
     - Frees decompressed `model.dat` buffer early before scene hierarchy construction.

## Tests & Benchmarks
- **Native Test Suite**: 218 / 218 tests passing (100%), including `Parser.StreamingDefinitionsPreservesGeometryAndHierarchy`.
- **WASM Benchmark**:
  - `SU_File.skp` (Legacy): **56.9 ms**
  - `capilla_quiroz_v17.skp` (2017): **18.0 ms**
  - `gondola_v20.skp` (2020): **466.9 ms**
  - `Untitled.skp` (Modern VFF, 16,949 faces): **491.2 ms** (623.8 KB ThatOpen Fragments output)
- **Backward Compatibility**: Native C++ API for `libOpenSkp` remains 100% backward-compatible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ __pycache__/
 # Distribution / packaging
 dist/
 build/
+packages/cpp/build-*/
 *.egg-info/
 *.egg
 .eggs/
--- a/packages/cpp/CMakeLists.txt
+++ b/packages/cpp/CMakeLists.txt
@@ -208,6 +208,10 @@ if(MSVC)
   target_compile_definitions(OpenSkp PRIVATE _CRT_SECURE_NO_WARNINGS)
 endif()
 
+if(OPENSKP_BUILD_WASM AND NOT EMSCRIPTEN)
+  message(FATAL_ERROR "Building WebAssembly target (OPENSKP_BUILD_WASM) requires the Emscripten toolchain (emcmake cmake ...).")
+endif()
+
 if(EMSCRIPTEN OR OPENSKP_BUILD_WASM)
   target_compile_options(OpenSkp PRIVATE -O3 -fexceptions)
   add_executable(openskp_wasm src/wasm_api.cpp)
@@ -234,7 +238,6 @@ if(EMSCRIPTEN OR OPENSKP_BUILD_WASM)
       "--bind"
       "-sWASM=1"
       "-sALLOW_MEMORY_GROWTH=1"
-      "-sMAXIMUM_MEMORY=4GB"
       "-sMODULARIZE=1"
       "-sEXPORT_ES6=1"
       "-sEXPORT_NAME=OpenSkpModule"
--- a/packages/cpp/CMakeLists.txt
+++ b/packages/cpp/CMakeLists.txt
@@ -26,6 +26,7 @@ endif()
 
 option(OPENSKP_BUILD_TESTS "Build OpenSkp tests" ${_openskp_tests_default})
 option(OPENSKP_BUILD_EXAMPLES "Build OpenSkp examples" OFF)
+option(OPENSKP_BUILD_WASM "Build OpenSkp WebAssembly / Emscripten module" OFF)
 
 # Developer tooling
 
@@ -205,6 +206,43 @@ endif()
 
 if(MSVC)
   target_compile_definitions(OpenSkp PRIVATE _CRT_SECURE_NO_WARNINGS)
+endif()
+
+if(EMSCRIPTEN OR OPENSKP_BUILD_WASM)
+  target_compile_options(OpenSkp PRIVATE -O3 -fexceptions)
+  add_executable(openskp_wasm src/wasm_api.cpp)
+  target_compile_options(openskp_wasm PRIVATE -O3 -fexceptions)
+  target_link_libraries(openskp_wasm PRIVATE OpenSkp)
+  target_include_directories(
+    openskp_wasm
+    PRIVATE
+      "${CMAKE_CURRENT_SOURCE_DIR}/src"
+      "${OPENSKP_MINIZ_INCLUDE_DIR}"
+      "${miniz_SOURCE_DIR}"
+      "${tinygltf_SOURCE_DIR}"
+      "${flatbuffers_SOURCE_DIR}/include"
+  )
+  set_target_properties(
+    openskp_wasm
+    PROPERTIES
+      OUTPUT_NAME "openskp"
+      SUFFIX ".js"
+  )
+  target_link_options(
+    openskp_wasm
+    PRIVATE
+      "--bind"
+      "-sWASM=1"
+      "-sALLOW_MEMORY_GROWTH=1"
+      "-sMAXIMUM_MEMORY=4GB"
+      "-sMODULARIZE=1"
+      "-sEXPORT_ES6=1"
+      "-sEXPORT_NAME=OpenSkpModule"
+      "-sENVIRONMENT=web,worker,node"
+      "-sDYNAMIC_EXECUTION=0"
+      "-fexceptions"
+      "-O3"
+  )
 endif()
 
 # Installation and CMake package
--- a/packages/cpp/build-wasm.bat
+++ b/packages/cpp/build-wasm.bat
@@ -1,0 +1,22 @@
+@echo off
+setlocal enabledelayedexpansion
+
+echo [OpenSKP] Building WebAssembly (WASM) module with Emscripten...
+
+if not exist build-wasm mkdir build-wasm
+cd build-wasm
+
+call emcmake cmake .. -DCMAKE_BUILD_TYPE=Release -DOPENSKP_BUILD_TESTS=OFF -DOPENSKP_BUILD_EXAMPLES=OFF -DOPENSKP_BUILD_WASM=ON
+if %ERRORLEVEL% neq 0 (
+    echo [OpenSKP] CMake configuration failed. Ensure emsdk is activated (e.g. emsdk_env.bat).
+    exit /b %ERRORLEVEL%
+)
+
+call emmake cmake --build . --config Release
+if %ERRORLEVEL% neq 0 (
+    echo [OpenSKP] Build failed.
+    exit /b %ERRORLEVEL%
+)
+
+echo [OpenSKP] WASM build successful! Output files in build-wasm/
+cd ..
--- a/packages/cpp/build-wasm.sh
+++ b/packages/cpp/build-wasm.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+set -e
+
+echo "[OpenSKP] Building WebAssembly (WASM) module with Emscripten..."
+
+mkdir -p build-wasm
+cd build-wasm
+
+emcmake cmake .. \
+  -DCMAKE_BUILD_TYPE=Release \
+  -DOPENSKP_BUILD_TESTS=OFF \
+  -DOPENSKP_BUILD_EXAMPLES=OFF \
+  -DOPENSKP_BUILD_WASM=ON
+
+emmake cmake --build . --config Release
+
+echo "[OpenSKP] WASM build successful! Output files in build-wasm/"
--- a/packages/cpp/src/core.cpp
+++ b/packages/cpp/src/core.cpp
@@ -1,5 +1,6 @@
 #include <algorithm>
 #include <cctype>
+#include <limits>
 #include <miniz.h>
 #include <regex>
 
@@ -61,6 +62,13 @@ struct Zip {
   static void validate_entry_size(const mz_zip_archive_file_stat& s) {
     auto declared = s.m_uncomp_size;
     if (declared == 0) return;
+
+    if (declared > std::numeric_limits<std::size_t>::max()) {
+      throw SkpParseError("ZIP entry '" + std::string(s.m_filename) + "' declares " +
+                              std::to_string(declared) +
+                              " bytes, exceeding addressable memory limits",
+                          ParseStage::zip_extract);
+    }
 
     if (declared > kMaxUncompressedEntryBytes) {
       throw SkpParseError("ZIP entry '" + std::string(s.m_filename) + "' declares " +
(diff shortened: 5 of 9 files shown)


iamahsanmehmood (changes requested):
Thanks for this — the WASM bindings and the streaming approach for large models are both genuinely useful additions. CI is green, but I found a few correctness issues in the streaming parser that I think need addressing before merge, the first one in particular.

## 1. `parse_tlv_recursive` now silently truncates any leaf payload over 64KB — globally, not just on the streaming path

`packages/cpp/src/tlv.cpp`:
```cpp
if (n.children.empty() && size && size <= 65536)
    n.payload.assign(...)
```
This function is the general-purpose TLV reader used by every parse path in the codebase, not just the new streaming one. Any leaf record whose payload is larger than 64KB now comes back with an **empty** `payload` instead of the actual bytes — silently, with no error and no log line. Nothing in the test suite exercises a >64KB leaf payload, so a real file with one (a large attribute-dictionary blob, embedded classification data, etc.) would quietly lose that data with no indication anything went wrong. If the intent is to bound memory only for the streaming/large-model path, this needs to be scoped there instead of applied to every caller of `parse_tlv_recursive`.

## 2. Dimensions silently dropped for files >50MB

`packages/cpp/src/core.cpp` skips `scan_vertex_positions`/`scan_instance_transforms` once `model->size() > 50MB`. Since `p.dimensions` is only computed `if (!vertex_positions.empty())`, large files now get no dimensions metadata at all, with nothing surfacing that this happened. Is that an accepted tradeoff? If so, worth a comment explaining why, plus a test on a >50MB fixture confirming the rest of the output (geometry/hierarchy) is still correct without dimensions.

## 3. Streaming definitions parser hand-duplicates the TLV traversal instead of reusing `parse_tlv_recursive`

The new lambda in `full_parse` manually walks F901 → 7017 → 7117 → 7C15 by hand rather than calling the existing recursive parser (the tag hierarchy does match the known container tags in `tlv.cpp`, so it's plausible). This is a second, independent implementation of the same binary-parsing logic — any future fix to the real recursive parser won't apply here. It also silently skips a definition when the inner `parse_tlv_recursive(...)` call comes back empty, unlike the pages/dimensions failures elsewhere in this same diff, which now log via `emit_log(..., LogLevel::debug, ...)`. Would rather see this reuse the existing recursive parser if at all possible, or at least log on the empty-parse case for consistency.

## 4. `respect_edge_visibility` is accepted but never used

Both `parseSkpToFragments` and `parseSkpToGLB` in `wasm_api.cpp` take a `respect_edge_visibility` bool that's never read anywhere — `ParseOptions` has no such field. Right now this silently does nothing regardless of what a JS caller passes, which is misleading for anyone integrating against this API. Either wire it up or drop it from the signature until it's implemented.

## 5. Header-prefix skip only in one of the two WASM entry points

`parse_skp_to_fragments` scans the first 512 bytes for the real SketchUp magic (`0xFF 0xFE 0xFF 0x0E` — matches `valid_header` in `vff.cpp`) and strips anything before it, but `parse_skp_to_glb` doesn't do the same thing. If a buffer handed in from JS can legitimately have extra bytes prepended before the real file content, both entry points need the same handling; if it can't, this is dead defensive code that could mask real corruption by truncating instead of failing loudly. Worth tracking down why this was needed in the first place.

## 6. The new test doesn't actually guard the streaming path

`Parser.StreamingDefinitionsPreservesGeometryAndHierarchy` only checks that geometry/hierarchy exist and face count is above a floor, on a fixture (`Untitled.skp`) that may not even be large enough to hit the streaming F901 branch. For a change this invasive to the core parser, I'd want a test that confirms output parity with the pre-change parser on the same file — same definitions/layers/materials/dimensions — not just "doesn't crash."

Happy to help figure out scoping for #1 if useful — that's the one I'd treat as blocking.

metehn:
Adding WebAssembly support and bringing the streaming F901 definition unwrap to C++ (resolving #264) is a great step forward for memory parity and web usage.
A few key findings and recommendations before merging:
1. **Safari/WebKit Memory Limit (`-sMAXIMUM_MEMORY=4GB`)**:
   WebKit/Safari (especially iOS Safari and macOS) will fail to instantiate the module with `RangeError: WebAssembly.Memory(): maximum memory size is larger than implementation limit` when set to 4GB. Reverting to Emscripten's default 2GB limit ensures cross-browser stability.
2. **CMake check for Emscripten**:
   `if(EMSCRIPTEN OR OPENSKP_BUILD_WASM)` causes issues when configuring with native toolchains using `-DOPENSKP_BUILD_WASM=ON`. Guard it with `if(OPENSKP_BUILD_WASM AND NOT EMSCRIPTEN) message(FATAL_ERROR ...)`.
3. **32-bit Integer Overflow Check in `Zip::get`**:
   `s.m_uncomp_size` is a 64-bit uint, while `std::size_t` is 32-bit on wasm32. Please ensure `s.m_uncomp_size <= std::numeric_limits<std::size_t>::max()` before casting.
4. **Progress Reporting in Streaming Loop**:
   `emit_progress(..., def_count, def_count + 1000)` produces inaccurate percentages and causes progress regression once returning to the outer loop. Consider mirroring the Python unwrap approach (`iter_top_level_lazy`).
5. **Input Validation in `wasm_api.cpp`**:
   Calling `uint8_array["byteLength"]` directly on `null`/`undefined` triggers an unhandled JS exception bypassing C++ `catch`. Adding `.isNull() || .isUndefined()` guards would make it much more robust.

nooraliqureshi:
Thanks for the feedback and suggestions @iamahsanmehmood and @metehn!

I've addressed both sets of review items across commits 4f36fc9 and e4a49d6:

### @iamahsanmehmood (Review feedback)

- **Leaf payload cap reverted (`tlv.cpp`)**: Removed the 64KB check so large leaf payloads are never truncated.
- **Dimensions preserved unconditionally (`core.cpp`)**: Removed the 50MB check; `scan_vertex_positions` and `scan_instance_transforms` now run for all files so dimension metadata is always retained.
- **TLV traversal reuse & logging (`core.cpp`)**: Reused `parse_tlv_recursive` for each definition record rather than duplicating traversal logic, and added debug logging on empty parse cases.
- **Unused argument removed (`wasm_api.cpp`)**: Dropped `respect_edge_visibility` from the WASM export functions and Emscripten bindings.
- **Magic scan prefix removed (`wasm_api.cpp`)**: Dropped the 512-byte scan from `parse_skp_to_fragments` so both WASM entry points fail consistently on invalid buffers.
- **Streaming test coverage (`parser_test.cpp`)**: Updated `StreamingDefinitionsPreservesGeometryAndHierarchy` with exact baseline assertions (46 definitions, 14 layers, 15 materials, 13 dimensions, 8 GLTF materials, 16,949 faces).

### @metehn (Recommendations)

- **Safari/WebKit memory limit**: Reverted `-sMAXIMUM_MEMORY=4GB` back to Emscripten's default 2GB ceiling to avoid `RangeError` instantiation failures on WebKit.
- **CMake Emscripten guard**: Added an explicit check in `CMakeLists.txt` so configuring `-DOPENSKP_BUILD_WASM=ON` with a native toolchain aborts with a clear fatal error.
- **32-bit integer overflow guard**: Added a bounds check against `std::numeric_limits<std::size_t>::max()` in `Zip::validate_entry_size` to protect against size truncation on wasm32.
- **Monotonic streaming progress**: Mirrored Python's `iter_top_level_lazy` and `_unwrap_definitions_container` by unwrapping `F901` definitions up front into the flat header list. This computes the exact total ahead of time and gives smooth, monotonic progress reporting without regression.
- **Missing input validation**: Added `.isNull() || .isUndefined()` guards in `wasm_api.cpp` to reject missing/null input cleanly before accessing buffer properties.

All 218 native unit tests and clang-format checks pass cleanly.

iamahsanmehmood (approved):
All 6 points from the earlier review are addressed, and re-verified rather than taken on faith:

1. **Silent 64KB payload truncation** - the cap in `tlv.cpp` is gone entirely; confirmed the diff against `main` shows no size-based truncation left anywhere in `parse_tlv_recursive`.
2. **Dimensions dropped for >50MB files** - the `model->size() <= 50MB` guard around `scan_vertex_positions`/`scan_instance_transforms` is gone; they now run unconditionally, and the new test asserts `model.dimensions.size() == 13` on a real fixture, so this is actually covered now, not just fixed.
3. **Hand-duplicated TLV traversal** - replaced with `unwrap_definitions_container()`, which reuses the existing `headers()` scan instead of a second hand-rolled recursive parser, and explicitly mirrors Python's own `_unwrap_definitions_container` (including its fall-back-to-treating-F901-as-an-ordinary-record behavior when the expected wrapper shape isn't found). Better than what I suggested, and it means this streaming path shares its correctness with an already-proven implementation instead of duplicating logic. Also now logs on an empty parse.
4. **Dead `respect_edge_visibility` parameter** - removed from both WASM-exposed functions.
5. **Inconsistent header-prefix skip** - removed entirely rather than duplicated, plus added explicit null/undefined input checks on both functions.
6. **Weak regression test** - now asserts exact counts (46 definitions, 14 layers, 15 materials, 13 dimensions, 8 materials, 16949 faces) against a real fixture instead of a vague floor.

Rebuilt and ran the full C++ suite on this branch directly: 218/218 passed, including the new precise test and all 5 real-fixture instanced-scene-parity tests. CI is green across the board including sanitizers.

Nice work turning this around - the F901-unwrap approach ended up cleaner than the original streaming lambda. Approving.

Issue #285, opened 2026-09-09
Cross-language porting backlog: bring TypeScript/.NET/Dart/C++ to parity with Python's newest features


iamahsanmehmood:
## Summary

Python has pulled ahead of the other 4 language ports (TypeScript, .NET, Dart, C++) on a number of real, shipped features. This is a tracking issue for bringing them to parity — one line per feature, not one issue per language. Check items off as PRs land; each checkbox links back to this issue when its PR closes it.

See [`docs/LANGUAGE_PARITY.md`](https://github.com/iamahsanmehmood/openskp/blob/main/docs/LANGUAGE_PARITY.md) for the full current-state matrix this list is derived from, and [`ROADMAP.md`](https://github.com/iamahsanmehmood/openskp/blob/main/ROADMAP.md) for the broader picture.

## Port Python → TypeScript / .NET / Dart / C++

- [ ] Attribute dictionaries: full 9-value-type support (`None`/`bool`/`Point3d`/`Vector3d`/`Length`/`Timestamp`/nested list-tuple). Other 4 languages: str/int/float only.
- [ ] Attribute dictionaries: multiple dictionaries per entity (`attribute_dicts=[...]`). Other 4: single dict only.
- [ ] Writer: linear dimensions (`add_dimension`) + leader text (`add_text`).
- [ ] Writer: section planes (`add_section_plane`).
- [ ] Writer + reader: construction lines/points.
- [ ] Legacy (pre-2021) pages/scenes reading (the other 4 languages only read pages/scenes from VFF/2021+ files).
- [x] Direct SKP→Fragments (.frag) export (`openskp.export.fragments`). **C++: done, merged** (`packages/cpp/src/fragments_export.cpp`). **TypeScript: done, merged** (#276) — canonical ThatOpen FlatBuffers schema, TRS decomposition and scale/mirror baking matching Python/C++, real-fixture tests, and verified via a real round-trip through the `@thatopen/fragments` npm package's own `SingleThreadedFragmentsModel`. **.NET/Dart: not started, no branch, no issue.**
- [ ] VFF per-layer-hidden flag reading (`8E3C` tag on `8C3C` layer nodes).
[...]

PR #304, opened 2026-09-10
fix(codegen): preserve applied_width and opacity for textured materials

metehn:
## Summary
Fixes a cross-language backlog item tracked in #285:
> codegen (`to_*_code()`) doesn't round-trip `applied_width`/opacity when regenerating a textured material â€” affects all 5 languages equally, mechanical fix, no ground-truth research needed.

When regenerating code from an existing model, textured materials previously omitted `applied_width` (in TypeScript, .NET, Dart, and C++) and `opacity` (across all 5 languages).

## Changes
- **Python (`openskp/codegen.py`)**: pass `opacity` when `mat.transparency < 1.0`.
- **TypeScript (`src/codegen.ts`)**: pass `appliedHeight`, `appliedWidth`, and optional `opacity`.
- **.NET (`OpenSkp/Codegen.cs`)**: pass `appliedHeight`, `appliedWidth`, and optional `opacity`.
- **Dart (`lib/src/codegen.dart`)**: pass `appliedHeight`, `appliedWidth`, and optional `opacity`.
- **C++ (`src/codegen.cpp`)**: pass `applied_height`, `applied_width`, and optional `opacity`.
- Added regression unit tests across all 5 packages asserting that `applied_width` and `opacity` round-trip accurately.
- Updated `docs/LANGUAGE_PARITY.md` and documented the fix under `CHANGELOG.md`'s `[Unreleased]` section.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed — Cross-language codegen textured material round-trip
+
+`to_*_code()` now preserves both `applied_width` and material `opacity` when regenerating textured materials across all 5 language ports (Python, TypeScript, .NET, Dart, C++).
+
 ### Changed — TypeScript writer memory
 
 `ArchiveWriter` keeps the archive in a growable `Uint8Array` instead of a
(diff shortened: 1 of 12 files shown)


iamahsanmehmood:
Thanks for the contribution and welcome! Taking a look now.

iamahsanmehmood (approved):
Nice work, and thanks for the clean cross-language fix - all 5 ports updated consistently, each with its own regression test.

Verified beyond CI: ran the Python test suite (547 passed) and the TypeScript codegen suite on this branch directly, and separately reproduced the fix's core claim by hand - generating code from a real textured material and confirming the regenerated `Texture.width`/`.height`/`transparency` now round-trip correctly instead of silently resetting to 1.0/opaque.

One thing I dug into while reviewing: I wanted to make sure passing the real applied width/height into the generated `addTextureMaterial` call couldn't disturb UV mapping for explicitly-pinned faces (since the old code's comment claimed 1.0 was a "hard requirement" there). I built a face with an explicit UV pin and a non-unit tile size and compared the regenerated UV transform against main - it's byte-identical before and after this PR, so this change doesn't touch that behavior at all. (There does appear to be a separate, pre-existing UV-transform scaling mismatch in that scenario, but it's already there on main independent of this PR, so it's not something this PR needs to fix - flagging separately.)

Approving.

PR #311, opened 2026-09-11
test(dotnet): implement SdkOracleFact to dynamically skip oracle tests in xUnit v2

metehn:
## Summary
Dynamically skip Trimble SketchUp SDK oracle tests in xUnit v2 using a custom [SdkOracleFact] discoverer when the native library is absent or fails to initialize.

Part of #285

## Type
- [x] Bug fix
- [ ] New feature
- [ ] Documentation

## Changes
- Replaced the soft-skip early return and unsupported `Assert.Skip` with custom `SdkOracleFactAttribute`, `SdkOracleFactDiscoverer`, and `SdkOracleTestCase`.
- Tests now report as native `Skipped` in xUnit test runners instead of false-positive `Passed`.
- Zero external package dependencies added.

## Testing
- .NET CI passed on Ubuntu, macOS, and Windows.
- Verified test suite skips cleanly on platforms without the native SDK DLL.

## Checklist
- [x] Tests pass locally
- [x] Documentation updated (if applicable)

### DIFF
--- a/packages/dotnet/OpenSkp.Tests/RealSketchUpOracleTests.cs
+++ b/packages/dotnet/OpenSkp.Tests/RealSketchUpOracleTests.cs
@@ -1,14 +1,81 @@
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text;
 using Xunit;
+using Xunit.Abstractions;
+using Xunit.Sdk;
 using OpenSkp;
 
 namespace OpenSkp.Tests
 {
+    /// <summary>Custom [Fact] attribute that dynamically skips SDK oracle tests
+    /// when the Trimble SketchUp SDK native library is absent or fails to initialize.</summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
+    [XunitTestCaseDiscoverer("OpenSkp.Tests.SdkOracleFactDiscoverer", "OpenSkp.Tests")]
+    public sealed class SdkOracleFactAttribute : FactAttribute
+    {
+    }
+
+    /// <summary>Discoverer for <see cref="SdkOracleFactAttribute"/>.</summary>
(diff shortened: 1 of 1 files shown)


iamahsanmehmood (approved):
Reviewed and verified locally, not just via CI.

**Design**: this is the correct, standard xUnit v2 pattern for a dynamically-determined skip reason — custom `[Fact]`-derived attribute + `FactDiscoverer` override + `XunitTestCase.GetSkipReason()` override. The `[Obsolete("Called by the de-serializer")]` parameterless constructor and the `#nullable disable/restore` scoping around it match what xUnit's own built-in test cases require for the same reason (deserialization support), and the build is genuinely warning-clean.

**Verified the actual fix, not just the code shape**: this dev machine happens to have a real SketchUp 2025 install with `SketchUpAPI.dll` present but `SUInitialize()` failing — literally the exact scenario the removed code comment described. Built and ran the oracle tests both before and after this PR on this machine:

- **Before** (`main`): `Passed! - Passed: 10, Skipped: 0` — a false positive, zero real assertions executed on any of the 10 tests.
- **After** (this PR): `Skipped OpenSkp.Tests.RealSketchUpOracleTests.* — Trimble SketchUp SDK (SketchUpAPI.dll) is not available or failed to initialize.` — all 10 correctly reported as skipped with a clear reason.

Also ran the full `OpenSkp.Tests` suite on this PR: 191 passed, 10 skipped, 0 failed — no regressions elsewhere.

**One small, non-blocking note**: the original constructor comment explaining exactly why `InvalidOperationException` specifically gets caught (a documented, reproduced `SUInitialize() -> 1` failure mode, cross-checked against plain ctypes from Python) got dropped when the catch widened to `catch (Exception)`. Broadening the catch is the right call — any initialization failure should skip, not just that one exception type — but that context was genuinely useful for whoever debugs this next and it'd be a shame to lose it. Not asking for a changed here, just flagging it in case you want to fold a short version of that note into the new code somewhere.

Nice work — this closes out the .NET item in #285's backlog. Approving.

PR #316, opened 2026-09-12
feat(python): loose-edge curve sets and analytic arcs, with four IFC4 conformance fixes

gsuiaoll:
## Summary

A drawing-style SketchUp model — a facade elevation, a section outline — is loose edges only, with no faces anywhere. `build_scene()` had nothing to bake, so the IFC export came out as a spatial skeleton with no elements in it at all: real line work, present in the `.skp`, absent from the `.ifc`. This adds those elements, and fixes four independent ways the IFC4 writer was emitting STEP that a conforming reader rejects.

Four values the parsers were reading and then discarding are now surfaced; loose edges are chained into runs along SketchUp's own `Edge#curve` grouping instead of a vertex-adjacency heuristic; and each run is written as an `IfcAnnotation` holding an `IfcGeometricCurveSet` in an `Annotation` sub-context, so the layer the author drew on survives into the file. A run that is provably one whole circular arc goes out as an `IfcIndexedPolyCurve` with `IfcArcIndex` segments rather than as its tessellation.

`base-commit: 3798b70b586cfcf7ca3c13f3a4207113932b2075` (current `main`, and what `git format-patch --base=auto` produces for this series).

## Type

- [x] Bug fix
- [x] New feature
- [ ] Documentation

Both boxes because the series is genuinely both: commits 1, 2 and 4 add the feature, commit 3 is a standalone bug fix that stands on its own.

## Changes

Five commits, each carrying its own tests and green on its own, so the series bisects:

| Commit | Subject | Test file it adds |
|:---|:---|:---|
| `09c7446` | surface edge layers, `Edge#curve` and analytic arc frames from both parsers | `test_edge_surfacing.py` (15 tests) |
| `9464882` | bake loose-edge runs into the scene, with analytic arcs | `test_loose_edge_curve_sets.py` (22 tests) |
| `80183ce` | emit valid IFC4 (STEP escaping, enum literals, per-type attribute counts, real `Closed`) | `test_ifc4_conformance.py` (13 tests) |
| `686a818` | export loose-edge curve sets as `IfcAnnotation` | `test_ifc_curve_annotations.py` (19 tests) |
| `4a1d3b4` | changelog + `docs/LANGUAGE_PARITY.md` for the above | — |

12 files, +3784/−30 in total: the four feature commits are 10 files, +3752/−30, and the docs commit adds `CHANGELOG.md` and `docs/LANGUAGE_PARITY.md`. Where the changes are:

- **`legacy.py` / `_core.py`** — the edge's own layer (VFF `D207`; the classic drawbase), the file's own `Edge#curve` (`BB0B`; the classic `CCurve` pointer), the 14 doubles of a classic `CArcCurve`, and `face['layer']`. On the face layer: the classic walker *did* parse the face drawbase and copied only material/hidden out of it, so every face arrived on whichever layer it happened to inherit — which is also why the "classify by layer" path never fired on real legacy files.
- **`scene.py`** — `_chain_loose_edges()` groups loose edges by the file's own curves. `_solve_run_arc()` calls a run an arc only when three things hold: it is the entire curve; its frame is a circle and not an ellipse (`|normal × x_axis − y_axis|` within tolerance); and every vertex lies on that circle within tolerance.
- **`export/ifc.py`** — the `IfcAnnotation` path, the four conformance fixes below, and `export/ifc_attr_counts.json`, generated from the schema rather than typed by hand.
- **`model.py`** — `build_scene(include_curve_sets=True)`, a new *defaulted* kwarg; cost is a lever (+5.7% of the IFC on a model with both solids and curves), not correctness. On a curve-only model there is nothing to opt out of.

**The four IFC4 defects** — all pre-existing, all found by reading what a real reader said about a real export:

1. non-ASCII text written as raw UTF-8, where ISO 10303-21 requires `\X2\<UTF-16BE hex>\X0\`;
2. IFC2x enum spellings — `.TRUE.`/`.FALSE.` are not IFC4 (`.T.`/`.F.` are), `.READWRITE.` is not a valid `IfcChangeActionEnum`, `.STERADIANUNIT.` is not a valid `IfcUnitEnum`;
3. every product line written with 9 attributes regardless of type — `IfcDoor` declares 13, `IfcWall` 9, so longer types silently lost their trailing attributes;
4. `IfcTriangulatedFaceSet.Closed` was a hardcoded `True`, now computed (welded at 6dp, every undirected edge used by exactly two triangles).

**Deliberate limits**, asserted in the tests rather than left implicit:

- The arc step bound carries a relative tolerance. A 4-sided SketchUp circle tessellates a closed run to `[π/2, π/2, π/2, π/2 + 2e-16]` (measured), and the closing step alone sat over a bare `π/2`, so the whole run was refused over float noise and fell back to its chords. The bound is a statement about evidence, not a measurement.
- A run that cannot be covered as at most two half-turn segments is conservatively left as chords. Guessing a segmentation is worse than the file's own chords, and IFC has no elliptical-arc expression in this shape at all — gondola's 16 non-uniformly-scaled arc frames out of 30 stay chords by design, not by omission.
- The VFF arc payload is not decoded: no file in the corpus carries an `ARC_CURVES` section to decode. That is asserted as a limit (`test_no_arc_frames_are_decoded`) so it cannot be mistaken for coverage; a future implementation should delete that test, not work around it.

## Testing

- `python -m pytest tests/` — **594 passed, 33 skipped** at the tip. Also run at each of the four feature commits (540 / 562 / 575 / 594 passed, zero failures at every one), so no commit in the series is broken.
- `ruff check packages/python/src/ packages/python/tests/` — clean, on the pinned `ruff==0.15.13`.
- **Mutation-tested, not merely green.** 24 mutations, each reverting one line of the fix, run one at a time: 24/24 killed, and each checked to fail *at the assertion that claims the behaviour*, not merely somewhere. Two survived on the first pass and both were worth the exercise: one was my mutation hitting only one of two write sites for the same value, the other was a real coverage hole (next bullet).
- **A real coverage hole, found and closed by that exercise.** No fixture in the corpus has a loose edge belonging to a `Curve` (verified across the whole corpus), so the parser-tables → scene seam was never exercised. `test_loose_edge_curve_sets.py` now drives that seam on the real parse tables with the faces removed.
- **Two tests are driven by synthetic element records** (`TestVffElementRecords`), because no file in the corpus carries a VFF `D207`: a table read over a whole real file cannot distinguish "nothing here is layered" from "the walker never looks".
- **No new fixtures.** Everything runs on the fixtures already in `tests/fixtures/`. Nothing to license, nothing added to the repo.
- **`ifcopenshell` is an optional test dependency only** (`pytest.importorskip`, so those tests skip where it is absent). One test opens the exported file with a real reader and counts what comes back, because the writer's own density check cannot see a reader that stops early. The library never imports it, so the core parser stays stdlib-only.
- **Conformance numbers re-measured on this tree**, not carried over from notes: re-exporting ten files (seven in-repo fixtures plus three real ones) with the four fixes reverted and then in place gives **360 `validate(express_rules=True)` problems → 0**. The non-ASCII case is measured the same way, since `validate()` cannot see it — the reverted writer emits 35 raw non-ASCII characters and 0 escapes for Chinese element names, the fixed one 0 and 19.

## Checklist

- [x] Tests pass locally
- [x] Documentation updated (if applicable) — `CHANGELOG.md` § Unreleased, and `docs/LANGUAGE_PARITY.md` § 2 and § 3
- [x] No breaking changes to the public API — additions only: the new `Scene.curve_sets` field and the defaulted `include_curve_sets` kwarg

---

One note on the branch history: it has been force-pushed since it was first opened. The earlier tip (`0b5cb11`) carried the same four changes, but the comments in it were written as fork-local notes (`LOCAL PATCH (2026-09-12): …` markers). Those markers are now removed and the surrounding comments stand as ordinary explanations, so a review of the earlier revision will not line up with this one.


iamahsanmehmood:
Thank you for this — genuinely one of the most thoroughly evidenced PRs this repo has received, and a first contribution at that. I went through the full diff, checked out the branch, ran the existing suite, and independently verified several of the claims rather than taking the write-up at face value. Sharing what I found so the record is complete.

**Independently reproduced (not just re-read from the PR body):**
- `_step_text`: fed it the exact two reference strings from your docstring (`测试项目`, `emoji😀墙`) and the output matches your quoted `ifcopenshell`-derived escapes byte-for-byte.
- The three IFC4 conformance bugs are real on `main` right now (confirmed by grep: `.READWRITE.`, `.STERADIANUNIT.`, hardcoded `.TRUE.` on `IfcTriangulatedFaceSet` all still present). Exporting `gondola_v20.skp` through current `main` and running `ifcopenshell.validate(express_rules=True)` reproduces the `IfcCorrectUnitAssignment` failure live — the `SOLIDANGLEUNIT` fix is not a theoretical concern.
- Exporting the same fixture through your branch and validating the output: **0 issues**, file opens cleanly (459,498 entities, no truncation).
- All 558 existing Python tests pass on your branch, `ruff check` is clean, and I approved the fork's gated CI runs — full matrix (macOS/Linux/Windows × 3.10-3.12) is green.

**One thing I'd like before merging:** this PR adds real complexity — the arc-recovery geometry in `_solve_run_arc`, the curve-chaining walk, the STEP escaping state machine, attribute-count padding, and the edge-welding closed-mesh check — and none of it has an automated test. The manual verification in the PR body is unusually rigorous (real files, cross-checked against `ifcopenshell`'s own serializer, measured tolerances), but that evidence doesn't live in CI, so a future refactor has nothing to catch a regression here. Could you add a handful of focused tests before this merges? Concretely, the highest-value ones would be:
1. `_step_text`/`sanitize_name` on the two unicode cases in your own docstring — cheap, exact, and I've already confirmed the expected output.
2. A synthetic `_solve_run_arc` case (a real circle vs. a non-uniformly-scaled ellipse) asserting the ellipse is correctly rejected.
3. A small `to_ifc` + `ifcopenshell.validate()` smoke test guarding the three enum fixes, so a regression here fails loudly instead of silently producing invalid STEP again.
4. One case in `_attr_counts`/the product-line writer for a non-9-attribute type (e.g. `IfcDoor`) confirming the arg list is padded to the right length.

Everything else looks ready. Really appreciate the depth here — the per-field forensic comments (measured byte offsets, cross-parser agreement tables, the `IfcArcIndex` inline-vs-entity-line gotcha) made this fast to verify rather than something I had to take on faith.

iamahsanmehmood:
Verified the update — you added exactly the four tests I asked for, and then some (69 new tests total across `test_edge_surfacing.py`, `test_ifc4_conformance.py`, `test_ifc_curve_annotations.py`, `test_loose_edge_curve_sets.py`). Went through them rather than just trusting the count:

- The unicode-escaping tests pin the exact byte output, including the property-set-key escaping case I hadn't even asked for.
- `test_an_elliptical_frame_is_refused` covers the non-uniform-scaling rejection directly, plus a good set of adjacent edge cases (float-boundary tolerance, split curves, off-circle vertices) I wouldn't have thought to ask for individually.
- On the `ifcopenshell` smoke test — you went a different route than I suggested (text/regex assertions pinning the exact STEP bytes, rather than a round-trip through `ifcopenshell.validate()`) and explained why in the module docstring: `ifcopenshell` isn't a dependency of this package, and the goal is pinning what this exporter writes, not a particular reader's opinion of it. That's the right call — better than what I asked for, not just different.
- The `IfcTriangulatedFaceSet.Closed` welding tests correctly distinguish "per-face vertices don't fool the check" from "an edge used three times isn't closed," which is exactly the failure mode described in the PR body.

Full suite: 627 passed (up from 558), `ruff check` clean, and CI is green across the full matrix (macOS/Linux/Windows × 3.10-3.12) after re-approving the fork's gated runs on the new commits.

The CHANGELOG and LANGUAGE_PARITY entries are similarly careful — measured 360→0 on real files rather than asserted, correctly scoped as Python-only, and linked to #285 for the other four ports to pick up. Nothing left to ask for. Merging.

PR #367, opened 2026-09-20
cpp: match SketchUp UV basis and pin scale in the writer

hk-archiwood:
## Summary

Python's writer already parameterized a face's texture matrix in SketchUp's normal-derived basis and scaled `front_uv`/`back_uv` pins from tiles into texture-inches. The C++ port did not.

- **Writer (`create.cpp`)** solved the matrix in a first-edge basis (`points[1] − points[0]`). That agrees with SketchUp only when that edge happens to run along `Z × n`; a square listed from a different corner, a downward face, or a faceted curve of small quads came out turned 90° or 180°.
- **Reader copies (`face_groups.cpp` / `edit.cpp` / `codegen.cpp`)** used a `1e-9` vertical snap and the `(X, −Y)` downward mirror. Near-horizontal float noise flipped the basis, and every underside read back upside-down.
- **Pin scale** was missing: callers pin in tiles, but SketchUp stores the matrix in texture-inches and divides by the material's applied size on read. A tile other than 1 in came out that many times too big (a 2 m water tile was 78.74× oversized).

C++ now matches Python on all four copies of the basis (`U = normalize(Z × n)`, `V = n × U`; horizontal `(X, Y)`; downward `(−X, +Y)`; tilt sine `< 1e-3` stays on the world axes), stores applied size per textured-material slot, and scales pins in both `SkpBuilder::add_face` and `ComponentDefinitionBuilder::add_face`.

## Files

- `packages/cpp/src/create.cpp` — SketchUp basis, `scale_pins`, applied-size map
- `packages/cpp/src/face_groups.cpp`, `edit.cpp`, `codegen.cpp` — same basis as Python `_face_groups.face_uv_basis`
- `packages/cpp/tests/create_test.cpp` — vertex-order matrix, downward `(−X, +Y)`, 10-inch tile 5× scale, tilted edge-aligned scale
- `CHANGELOG.md` — Unreleased

## Test plan

- [x] `openskp_tests --gtest_filter='Create.*'` (57 tests, including the four new UV cases)
- [x] Full `openskp_tests` (250 tests)
- [ ] CI C++ format + build + sanitizers

Made with [Cursor](https://cursor.com)


iamahsanmehmood (dismissed):
Deep-reviewed against Python's actual ground truth, not just the PR's own description. `face_uv_basis`'s `U = normalize(Z × n), V = n × U`, the `VERTICAL_TOLERANCE = 1e-3` snap, and the `(-X, +Y)` downward basis all match `packages/python/src/openskp/_face_groups.py` byte-for-byte, including the exact tolerance value and its SDK-measurement provenance comment. The applied-size pin-scaling logic matches `create.py`'s `_scale_pins`/`_uv_matrix_for_face` too. This is a real, previously-undetected divergence from Python (and from real SketchUp) that the old first-edge-basis approach only happened to get right when a face's first edge ran along Z×n.

Verified independently, not just by reading the diff: built locally and ran the 4 new regression tests (vertex-order, downward-face, tile-scale, tilted-edge) plus the full suite — 249/250 passing, the 1 failure being the pre-existing, machine-load-sensitive `InstancedGlb` perf flake unrelated to this change. `clang-format` needed a small whitespace fix, which I pushed directly to this branch (maintainer edits enabled).

Solid work — merging.

iamahsanmehmood (approved):
Deep-reviewed against Python's actual ground truth, not just the PR's own description. `face_uv_basis`'s `U = normalize(Z × n), V = n × U`, the `VERTICAL_TOLERANCE = 1e-3` snap, and the `(-X, +Y)` downward basis all match `packages/python/src/openskp/_face_groups.py` byte-for-byte, including the exact tolerance value and its SDK-measurement provenance comment. The applied-size pin-scaling logic matches `create.py`'s `_scale_pins`/`_uv_matrix_for_face` too. This is a real, previously-undetected divergence from Python (and from real SketchUp) that the old first-edge-basis approach only happened to get right when a face's first edge ran along Z×n.

Verified independently, not just by reading the diff: built locally and ran the 4 new regression tests (vertex-order, downward-face, tile-scale, tilted-edge) plus the full suite — 249/250 passing, the 1 failure being the pre-existing, machine-load-sensitive `InstancedGlb` perf flake unrelated to this change. `clang-format` needed a small whitespace fix, which I pushed directly to this branch (maintainer edits enabled), and then merged current `main` in since the branch had fallen behind.

Solid work — merging.

PR #368, opened 2026-09-20
cpp: keep native types on instance attribute dictionaries

hk-archiwood:
## Summary
- Python's `Instance.attribute_dictionaries` already returns the types SketchUp stored (`int` / `float` / `None` / 3-tuple / nested list). C++ decoded the same VFF tags (`A738` int32, `A938`/`AF38` float64, `B438`/`B538` point/vector, `AE38` array) then immediately formatted them as strings, so a saved integer came back as `"-7"`.
- `Instance::attribute_dictionaries` is now `ParsedAttrDictionaries`. Every named dictionary is kept; the reader does not skip names such as `dynamic_attributes` or `SU_InstanceSet`. `properties` remains the existing stringified view of a dictionary named `dynamic_attributes` when that name is present.
- Scene / JSON / IFC stringify every dictionary. Legacy `CAttributeNamed` keeps the same native types, and non-DC dictionaries now actually land on the instance (they were dropped before).

## Test plan
- [x] `openskp_tests` — 246/246 passed
- [x] VFF geometry tests: integer / float / point / array / null
- [x] Writer round-trip: `Create.InstanceAttributesRoundTrip` asserts int32 `42`, double `3.5`, string `"widget"`
- [x] Named dictionaries including `dynamic_attributes` / `SU_InstanceSet` stay on the instance
- [x] Real-fixture `Parser.*` still pass


iamahsanmehmood:
Deep-reviewed this against Python's actual ground truth. The core claim checks out exactly: `Instance.attribute_dictionaries` (`model.py:364`, `_core.py:1105-1127`) really does keep native types (`int`/`float`/`None`/3-tuple/list), and the old C++ behavior of stringifying at decode time — plus silently excluding `dynamic_attributes`/`SU_InstanceSet` from `attribute_dicts` — was a real, previously-undetected divergence from that contract. `ParsedAttribute`/`ParsedAttrDictionaries` is a clean way to carry that. Nice catch, and nice test coverage on the typed-decode side.

One real issue I found that needs fixing before this merges, though: the fix removes the `dynamic_attributes`/`SU_InstanceSet` exclusion at the single point where it used to live (`geometry.cpp`'s `DC05` handling), but `RawInstance::attribute_dicts` feeds **three** different downstream consumers, and only one of them (the model-level `Instance::attribute_dictionaries`) is supposed to lose that filter. The other two — `InstancedNode::attribute_dictionaries` in `instanced_scene.cpp` and `InstanceNode::attribute_dictionaries` in `scene.cpp`, plus both files' name-override search loops (`for (const auto& dict : i.attribute_dicts) { ... "name"/"label"/"code" ... }`) — now silently pick up `dynamic_attributes` and `SU_InstanceSet` too, where they shouldn't.

Python keeps these as two genuinely different fields with two different filtering rules, not one shared source:
- `model.py`'s `Instance.attribute_dictionaries` (`_core.py:1105`, `inst_attribute_dicts = extract_attribute_dictionaries(d007)`) — unconditional, everything included. This is what your fix correctly targets.
- `scene.py`'s per-node `attribute_dicts` (`scene.py:642-644`, same logic in `instanced_scene.py:635`) — explicitly `if dict_name in ("dynamic_attributes", "SU_InstanceSet"): continue`, both for what lands in the node's own `attribute_dictionaries` **and** for what's eligible as a name-override candidate. The comment there is explicit about why: `SU_InstanceSet` is "SketchUp's own always-present, always-empty Owner/Status boilerplate, not worth surfacing," and `dynamic_attributes` is already exposed via `properties` so it'd otherwise be duplicated.

Concretely, with this PR as-is: a real file where `dynamic_attributes` happens to carry a `"name"` key (plausible for Dynamic Components) would now change that instance's resolved display name in the scene output, and `SU_InstanceSet`/`dynamic_attributes` would start showing up as extra (redundant/noise) entries in `InstancedNode.attribute_dictionaries`/`InstanceNode.attribute_dicts` — neither of which Python does.

Suggested fix: keep `geometry.cpp`'s `DC05` extraction unconditional (feeding the model-level field correctly, as you have it), but have `instanced_scene.cpp`/`scene.cpp` filter `dynamic_attributes`/`SU_InstanceSet` out of `i.attribute_dicts` locally before building `node.attribute_dictionaries` and before the name-override search loop — mirroring `scene.py`/`instanced_scene.py`'s own two-tier structure instead of relying on a single upstream filter. Happy to merge once that's addressed — the typed-value work itself is solid.

hk-archiwood:
Addressed in ff0b4e1.

`geometry.cpp` still extracts every dictionary into `RawInstance::attribute_dicts`, so model-level `Instance::attribute_dictionaries` stays unfiltered and typed. `scene.cpp` and `instanced_scene.cpp` now skip `dynamic_attributes` and `SU_InstanceSet` locally, both when filling the node dictionaries and when searching `name`/`label`/`code`. A `dynamic_attributes` `name` no longer changes the scene display name; plugin dictionaries still do.

iamahsanmehmood (dismissed):
Fix confirmed correct and complete now. To recap for the record: this PR went through two review rounds — first the scene/model attribute-dictionary filtering fix (verified, addressed cleanly in ff0b4e1), then a real cross-platform build failure that only showed up once merged against current `main`: `extract_legacy_attribute_dictionaries` collided with an identically-named, identically-signatured function #369 had already added for CLayer dictionaries (same params, different return type — not valid C++ overloading, and MSVC silently picked one while GCC correctly rejected it as ambiguous). I merged main in, found and fixed that collision myself (removed the duplicate, had the CLayer call site use the shared typed extractor + `stringify_attr_dictionaries()`), verified locally, and pushed directly to this branch.

CI is now fully green across every platform, including the ones that caught the real bug (ubuntu-clang, ubuntu-gcc, macos-clang all failed before the fix, all pass now). Full local suite: 257/257.

Merging.

iamahsanmehmood (approved):
Re-approving after the branch-update pushes (main had advanced twice more since my last review, both clean merges with no further code changes needed). CI fully green on the final commit. Merging.

PR #369, opened 2026-09-20
cpp: round-trip named dictionaries on CLayer

hk-archiwood:
## Summary

`SkpBuilder::add_layer` already had the `CAttributeNamed` writer used by faces and instances, but `CLayer` never kept its preamble and `Layer` had no dictionary slot. A caller that wrote extra dictionaries on a layer got a file SketchUp could open, then `parse()` dropped them.

This keeps the API generic:

- `LayerOptions::extra_dictionaries` writes named `CAttributeNamed` records on the layer.
- `parse()` copies every named dictionary on a legacy `CLayer` onto `Layer::attribute_dictionaries` (string maps, same contract as `Instance::attribute_dictionaries`).
- The reader does not special-case dictionary names.
- `edit()` replays the dictionaries so a round-trip through the writer keeps them.

VFF layers still have no attribute slot, so this stays empty there. Python `create.py` is unchanged (layers there are still color/hidden only).

## Test plan

- [x] `Create.LayerExtraDictionariesRoundTrip` writes `plugin`/`n` = int32 7 and reads `"7"` back
- [x] Full C++ suite: 247/247 passing locally (`openskp_tests`)
- [ ] CI C++ format + build on this PR

Made with [Cursor](https://cursor.com)


iamahsanmehmood (approved):
Deep-reviewed and independently verified locally: full C++ suite (247 tests) passes with the fix applied (the 1 failure, InstancedGlb's wall-clock perf test, is a pre-existing, machine-load-sensitive flake unrelated to this change), and `clang-format` needed a small whitespace fix which I pushed directly to this branch (maintainer edits are enabled) since it was purely mechanical.

The change is narrowly scoped, follows the existing `Instance::attribute_dictionaries` pattern exactly, and is honest about scope (Python's writer stays root-level-only, clearly stated rather than silently expanded). CI is fully green. Thanks for this — merging.

